Issue #27, opened 2023-04-11
improve return k8s from state


rawdaGastan:
we can try to load all fields of a k8s cluster

PR #79, opened 2023-04-16
k8s loader

rawdaGastan:
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/27

### DIFF
--- a/grid-client/state/state.go
+++ b/grid-client/state/state.go
@@ -1,5 +1,5 @@
-// Package deployer for grid deployer
-package deployer
+// Package state for grid state
+package state
 
 import (
 	"context"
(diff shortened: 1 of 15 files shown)


grid-client/state/state.go
```diff
@@ -181,6 +180,16 @@ func (st *State) LoadK8sFromGrid(nodeIDs []uint32, deploymentName string) (workl
 		return workloads.K8sCluster{}, fmt.Errorf("failed to get master node for k8s cluster %s", deploymentName)
 	}
 	cluster.NodeDeploymentID = nodeDeploymentID
+	cluster.NetworkName = cluster.Master.NetworkName
+	cluster.SSHKey = cluster.Master.SSHKey
+	cluster.Token = cluster.Master.Token
+
+	// get cluster IP ranges
+	err := st.AssignNodesIPRange(&cluster)
```
AbdelrahmanElawady, line 188: in order to load k8s we will need to have the cluster network already stored in state.
I think we can get the cluster network since we have its name and which nodes it is deployed on, that way we wouldn't depend on it being stored in state.

rawdaGastan: good idea 🎆  added to network issue

Issue #107, opened 2023-04-26
add network workload metadata


mariobassem:
to be able to rebuild networks without having state, we need to have some information in network workloads. 
information like:
- chosen public node
- user access private key
- AddWGAcess flag (could be deduced from network workloads, but would be nice to have)

rawdaGastan:
isn't it the same as https://github.com/threefoldtech/tfgrid-sdk-go/issues/29 ? 👀

PR #109, opened 2023-04-26
adds network workload metadata

mariobassem:
## Related issues:

- #107

### DIFF
--- a/grid-client/state/state.go
+++ b/grid-client/state/state.go
@@ -267,7 +267,6 @@ func (st *State) computeK8sDeploymentResources(nodeID uint32, dl gridtypes.Deplo
 func (st *State) LoadNetworkFromGrid(name string) (znet workloads.ZNet, err error) {
 	var zNets []workloads.ZNet
 	nodeDeploymentsIDs := map[uint32]uint64{}
-	metdata := workloads.NetworkMetaData{}
 
 	sub := st.Substrate
 	for nodeID := range st.CurrentNodeNetworks {
@@ -297,12 +296,6 @@ func (st *State) LoadNetworkFromGrid(name string) (znet workloads.ZNet, err erro
 						return workloads.ZNet{}, errors.Wrapf(err, "failed to get network from workload %s", name)
 					}
 
-					if metdata.PublicNodeID == 0 {
-						if err := json.Unmarshal([]byte(wl.Metadata), &metdata); err != nil {
-							return workloads.ZNet{}, errors.Wrapf(err, "failed to parse network metadata from workload %s", name)
-						}
-					}
-
 					break
 				}
 			}
@@ -323,6 +316,10 @@ func (st *State) LoadNetworkFromGrid(name string) (znet workloads.ZNet, err erro
 		maps.Copy(wgPort, net.WGPort)
 		maps.Copy(keys, net.Keys)
 		nodes = append(nodes, net.Nodes...)
+		znet.AccessWGConfig = net.AccessWGConfig
+		znet.ExternalIP = net.ExternalIP
+		znet.ExternalSK = net.ExternalSK
+		znet.PublicNodeID = net.PublicNodeID
 	}
 
 	znet.NodeDeploymentID = nodeDeploymentsIDs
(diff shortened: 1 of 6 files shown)


grid-client/workloads/network.go
```diff
@@ -24,6 +24,13 @@ type UserAccess struct {
 	PublicNodeEndpoint string
```
AbdelrahmanElawady, line 24: `NewNetworkFromWorkload` also needs to be updated to set metadata stored in the workload to znet struct, right?

mariobassem: Sure it should. I didn't add it because i think the znet retrieval in the grid client right now is somehow redundant. It retrieves information from only one workload, which is not entirely right i think.
Will update this, it would be really helpful.
Thanks

grid-client/state/state.go
```diff
@@ -295,6 +296,13 @@ func (st *State) LoadNetworkFromGrid(name string) (znet workloads.ZNet, err erro
 					if err != nil {
 						return workloads.ZNet{}, errors.Wrapf(err, "failed to get network from workload %s", name)
 					}
+
+					if metdata.PublicNodeID == 0 {
+						if err := json.Unmarshal([]byte(wl.Metadata), &metdata); err != nil {
```
rawdaGastan: I think we can implement metadat part in `NewNetworkFromWorkload`

grid-client/state/state.go
```diff
@@ -322,6 +330,24 @@ func (st *State) LoadNetworkFromGrid(name string) (znet workloads.ZNet, err erro
 	znet.NodesIPRange = nodesIPRange
 	znet.Keys = keys
 	znet.WGPort = wgPort
+	if metdata.UserAcessIP != "" {
```
rawdaGastan: this also should be moved to `NewNetworkFromWorkload`

grid-client/state/state.go
```diff
@@ -322,6 +330,24 @@ func (st *State) LoadNetworkFromGrid(name string) (znet workloads.ZNet, err erro
...
+	if metdata.PrivateKey != "" {
+		key, err := wgtypes.ParseKey(metdata.PrivateKey)
+		if err != nil {
+			return znet, errors.Wrapf(err, "failed to parse user access private key from metadata string %s", metdata.PrivateKey)
+		}
+		znet.ExternalSK = key
+	}
+
+	znet.PublicNodeID = metdata.PublicNodeID
+	znet.AddWGAccess = (metdata.PrivateKey != "")
```
rawdaGastan: I think you mentioned that we can't guarantee that AddWGAccess is true from the private key.
also we have the private key in the zos workload

rawdaGastan (changes requested):

grid-client/state/state.go
mariobassem: it could, but as i understand, the NewNetworkFromWorkload method gathers the relevant information per network workload, and the PublicNodeID is related to the whole network, so it would work yes, but we would be reassigning the same value over and over again for all network workloads.

grid-client/state/state.go
mariobassem: same as above

grid-client/state/state.go
mariobassem: this is correct for private keys of network workloads, you can't conclude that a user requested wg access from the presence of a private key in a network workloads, because it's always there. but the private key stored in metadata is a different one, it's the key assigned to a user whenever they request wg access. so basically, the presence of a private key or an ip in the metadata indicates that the user requested wg access before.

grid-client/workloads/network.go
rawdaGastan: I think it can be parsed from a zos workload so the functions should do it

grid-client/state/state.go
rawdaGastan: I think it can be parsed from a zos workload so the functions should do it

grid-client/state/state.go
```diff
@@ -315,6 +316,10 @@ func (st *State) LoadNetworkFromGrid(name string) (znet workloads.ZNet, err erro
 		maps.Copy(wgPort, net.WGPort)
 		maps.Copy(keys, net.Keys)
 		nodes = append(nodes, net.Nodes...)
+		znet.AccessWGConfig = net.AccessWGConfig
```
rawdaGastan: I think those 4 fields are assigned to znet in line 291?

rawdaGastan (changes requested):

rawdaGastan (approved):

Issue #116, opened 2023-04-30
add batch calls to the client (cancel and create node contracts)


rawdaGastan:
wait for https://github.com/threefoldtech/tfchain/issues/687

PR #142, opened 2023-05-14
make deployment process concurrent

mariobassem:
## Related issues:

- #174

### DIFF
--- a/grid-client/deployer/deployer.go
+++ b/grid-client/deployer/deployer.go
@@ -139,6 +139,10 @@ func (d *Deployer) deleteHandler(ctx context.Context, oldDls map[uint32]uint64, 
 		}
 	}
 
+	if len(contractsToDelete) == 0 {
+		return nil
+	}
+
 	if err := d.substrateConn.BatchCancelContract(d.identity, contractsToDelete); err != nil {
 		return err
 	}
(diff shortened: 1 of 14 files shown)


grid-client/deployer/deployer.go
```diff
@@ -155,35 +156,37 @@ func (d *Deployer) deploy(
...
 			log.Debug().Uint32("Number of public ips", publicIPCount)
 
-			contractID, err := d.substrateConn.CreateNodeContract(d.identity, node, dl.Metadata, hashHex, publicIPCount, newDeploymentSolutionProvider[node])
-			log.Debug().Uint64("CreateNodeContract returned id", contractID)
-			if err != nil {
-				return currentDeployments, errors.Wrap(err, "failed to create contract")
-			}
+			nodeID := node
+			dep := dl
+			errGroup.Go(func() error {
```
xmonader, line 195: I don't think this would work, as you need to maintain the sequence number update or use batch calls?

mariobassem: i tested this multiple times manually, and it was working fine...
but using batch calls seems more fun :D
but i think would require some, not so minor, changes in the current deployment process

mariobassem:
deletions and creations work fine, updates operation still needs some fixes.
also, tests don't pass, should have a look into it.

xmonader:
blocking comment for @rawdaGastan and @AbdelrahmanElawady

rawdaGastan:
We now support batch calls for deployments. So, we won't support concurrency for them. The user can use batch calls.
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/116

Issue #144, opened 2023-05-15
calculating deployment used ips is incorrect


mariobassem:
when deploying a deployment, a process of assigning IPs to VMs is carried. this process first tries to calculate the already used ips, then tries to assign unused IPs to VMs that still don't have an IP assigned.
the problem happens when reading already assigned IPs and parsing them, the net package by default parses IPs as IPv6, but the IPs we are using are IPv4, which should be handled accordingly, or be specifically parses as IPv4.

PR #145, opened 2023-05-15
Development fix network

mariobassem:
## Related issues

- #144

### DIFF
--- a/grid-client/state/network_state.go
+++ b/grid-client/state/network_state.go
@@ -35,13 +35,14 @@ func (nm NetworkState) GetNetwork(networkName string) Network {
 	return net
 }
 
-// UpdateNetwork updates a network subnets given its name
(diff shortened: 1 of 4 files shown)


grid-client/state/network_state.go
```diff
@@ -44,6 +44,16 @@ func (nm NetworkState) UpdateNetwork(networkName string, ipRange map[uint32]grid
 	}
 }
 
+// UpdateNetworkSubnets updates a network subnets given its name
+func (nm NetworkState) UpdateNetworkSubnets(networkName string, ipRange map[uint32]gridtypes.IPNet) {
```
rawdaGastan, line 39: why creating a new func? I think we can do it in `UpdateNetwork` func only

rawdaGastan (changes requested):
please remove the first commit as it belongs to another pr

rawdaGastan (approved):

PR #161, opened 2023-06-01
remove error check for "ContractNotExists" when canceling a contract

mariobassem:
## Description:

Currently, when a user cancels their contract, the subi interface checks whether the returned error string is "ContractNotExists" and if so, ignores it. but actually the returned error string is not just "ContractNotExists", the error includes it, so the check does not do anything.
And as stated below, this maybe the way to go, where we let the user decide what to do when such error returns. So, this pr removes the current check, which does not do anything except remove the confusion.

### DIFF
--- a/grid-client/subi/substrate_manager.go
+++ b/grid-client/subi/substrate_manager.go
@@ -3,6 +3,7 @@ package subi
 
 import (
 	"context"
+	"strings"
 	"sync"
 
 	"github.com/pkg/errors"
@@ -149,7 +150,7 @@ func (s *SubstrateImpl) CancelContract(identity substrate.Identity, contractID u
 	s.m.Lock()
 	defer s.m.Unlock()
 
-	if err := s.Substrate.CancelContract(identity, contractID); err != nil && err.Error() != "ContractNotExists" {
+	if err := s.Substrate.CancelContract(identity, contractID); err != nil && !strings.Contains(err.Error(), "ContractNotExists") {
 		return normalizeNotFoundErrors(err)
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


grid-client/subi/substrate_manager.go
```diff
@@ -149,7 +150,7 @@ func (s *SubstrateImpl) CancelContract(identity substrate.Identity, contractID u
 	s.m.Lock()
 	defer s.m.Unlock()
 
-	if err := s.Substrate.CancelContract(identity, contractID); err != nil && err.Error() != "ContractNotExists" {
+	if err := s.Substrate.CancelContract(identity, contractID); err != nil && !strings.Contains(err.Error(), "ContractNotExists") {
```
xmonader: isn't there a better way to check that error? like with Is or number constant comparison?

mariobassem: with Is, we can't, because this error is not exposed from the tfchain client, it's only a string (https://github.com/threefoldtech/tfchain/blob/development/clients/tfchain-client-go/utils.go#L58)

mariobassem: don't know what you mean with number constant comparison :(

xmonader: Alright, however I think it's the responsibility of the caller to check if the contract exists or not before deletion instead of introducing special handling in the code itself (even if for Idempotence)

DylanVerstraete:
Why are we adding this check? In what situation would you want to ignore this error?

grid-client/subi/substrate_manager.go
mariobassem: sure, will remove it

mariobassem:
because if the purpose is to cancel some contract, then i don't really see why a user would care if the contract has existed before the cancellation or not, the eventual state is the same i mean. but then again maybe this is useful to ensure no logic flaws for users.
will remove it anyways

alichaddad (approved):

Issue #175, opened 2023-06-08
gridproxy report zero contracts for anode listed in dedicated nodes although it has one active contract 


hossnys:
## Describe the bug

on testnet node 14 which is dedicated node , it's not able to rent the node but it's listed in rentable nodes , ans is reported by gridproxy that it has 0 deployments and 0 workloads
https://github.com/threefoldtech/tf_operations/issues/1724
https://user-images.githubusercontent.com/109473873/244090433-16e7e9a7-54ba-499b-aba4-4cc0d54d7b0c.webm

## Expected behavior

node shouldn't be listed in rentable table as long as there is an active contract on it .

## Screenshots
![Screenshot from 2023-06-08 14-08-34](https://github.com/threefoldtech/tfgrid-sdk-go/assets/26298009/b599dbd9-6e30-4ca1-bef6-332d1ffe326e)

![Firefox_Screenshot_2023-06-08T11-16-45 832Z](https://github.com/threefoldtech/tfgrid-sdk-go/assets/26298009/d5e2f23a-405d-4de7-87b6-f997af23cd0e)

khaledyoussef24:
verified and tested on devnet on many nodes.
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/19999704-c056-421c-8e6d-920ede6b7d84)

PR #176, opened 2023-06-08
fix rentable query

Omarabdul3ziz:
### Description 
- add join between node_contract and nodes tables distinct to prevent multiple rows for the same node.
- modify rentable query to check for a single node_contract associated with a node.

### Related Issue 
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/175

### DIFF
--- a/grid-proxy/internal/explorer/db/postgres.go
+++ b/grid-proxy/internal/explorer/db/postgres.go
@@ -332,6 +332,10 @@ func (d *PostgresDatabase) farmTableQuery() *gorm.DB {
 		)
 }
 func (d *PostgresDatabase) nodeTableQuery() *gorm.DB {
+	subquery := d.gormDB.Table("node_contract").
+		Select("DISTINCT ON (node_id) node_id, contract_id").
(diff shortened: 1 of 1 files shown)


alichaddad (commented):
I think this looks good, can you add a test case for this?

mariobassem (approved):

mariobassem:
lgtm, and if no one has comments, i guess we should remove the states shit here if not needed elsewhere:
https://github.com/threefoldtech/tfgrid-sdk-go/blob/e5cd5c0556056a1f0a0e33d51c97ac3b749179fb/grid-proxy/internal/explorer/db/postgres.go#L53-L56

Omarabdul3ziz:
@alichaddad the rentable filter already included in node tests here 
https://github.com/threefoldtech/tfgrid-sdk-go/blob/f0414098e80c6eb32443156c6cfe600078bdae87/grid-proxy/tests/queries/local_client.go#L453-L457

@MarioBassem yes, it is not used anywhere else. it is saved to remove.

Issue #210, opened 2023-06-20
update stats endpoint for gpus count


xmonader:
https://gridproxy.dev.grid.tf/stats

should include number of gpus

not sure if should be number of gpus or number of nodes with gpus enabled, given the node can have multiple GPUs

In case of nodes that has gpu enabled it's a simple query, but in case of the exact number of gpus i believe we will have to count all gpus using listGPUs

A-Harby:
Verified on Devnet,


![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/d142a250-0be2-4b90-a53a-4a901b430c6c)

PR #211, opened 2023-06-21
replace has gpu with num gpu

alichaddad:
### Changes

- return 1 as gpu count if node has gpu
- Count number of nodes with gpu and add it to the stats endpoint

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-go/issues/210

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/grid-proxy/internal/explorer/converters.go
+++ b/grid-proxy/internal/explorer/converters.go
@@ -26,6 +26,14 @@ func decideNodeStatus(power types.NodePower, updatedAt int64) string {
 	}
 }
 
+// getNumGPUs should be deleted after removing hasGPU
+func getNumGPUs(hasGPU bool) int {
+	if hasGPU {
+		return 1
+	}
+	return 0
+}
+
 func nodeFromDBNode(info db.Node) types.Node {
 	node := types.Node{
 		ID:              info.ID,
@@ -70,7 +78,7 @@ func nodeFromDBNode(info db.Node) types.Node {
 		RentedByTwinID:    uint(info.RentedByTwinID),
 		SerialNumber:      info.SerialNumber,
 		Power:             types.NodePower(info.Power),
-		HasGPU:            info.HasGPU,
+		NumGPU:            getNumGPUs(info.HasGPU),
 		ExtraFee:          info.ExtraFee,
 	}
 	node.Status = decideNodeStatus(node.Power, node.UpdatedAt)
@@ -141,7 +149,7 @@ func nodeWithNestedCapacityFromDBNode(info db.Node) types.NodeWithNestedCapacity
 		RentedByTwinID:    uint(info.RentedByTwinID),
 		SerialNumber:      info.SerialNumber,
 		Power:             types.NodePower(info.Power),
-		HasGPU:            info.HasGPU,
+		NumGPU:            getNumGPUs(info.HasGPU),
 		ExtraFee:          info.ExtraFee,
 	}
 	node.Status = decideNodeStatus(node.Power, node.UpdatedAt)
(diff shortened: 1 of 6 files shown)


grid-proxy/internal/explorer/converters.go
```diff
@@ -26,6 +26,16 @@ func decideNodeStatus(power types.NodePower, updatedAt int64) string {
 	}
 }
 
+// getNumGPUs should be deleted after removing hasGPU
+func getNumGPUs(hasGPU bool) int {
```
xmonader, line 30: let's do 
if hasGPU{ return 1} 
return 0

grid-proxy/internal/explorer/converters.go
```diff
@@ -70,7 +80,7 @@ func nodeFromDBNode(info db.Node) types.Node {
 		RentedByTwinID:    uint(info.RentedByTwinID),
 		SerialNumber:      info.SerialNumber,
 		Power:             types.NodePower(info.Power),
-		HasGPU:            info.HasGPU,
+		NumGPUs:           getNumGPUs(info.HasGPU),
```
xmonader: should be NumGPU not NumGPUs similar to runtime.NumCPU https://pkg.go.dev/runtime#NumCPU

grid-proxy/pkg/types/types.go
```diff
@@ -29,6 +29,7 @@ type Counters struct {
 	Twins             int64            `json:"twins"`
 	Contracts         int64            `json:"contracts"`
 	NodesDistribution map[string]int64 `json:"nodesDistribution" gorm:"-:all"`
+	GPUs              int64            `json:"gpus"`
```
xmonader, line 32: should we unify the json to `num_gpu` here and in 191?

grid-proxy/tests/queries/local_client.go
```diff
@@ -343,12 +343,22 @@ func (g *GridProxyClientimpl) Node(nodeID uint32) (res proxytypes.NodeWithNested
...
 		},
-		HasGPU:   node.HasGPU,
+		NumGPUs:  getNumGPUs(node.HasGPU),
 		ExtraFee: node.ExtraFee,
 	}
 	return
 }
 
+// getNumGPUs should be deleted after removing hasGPU
+func getNumGPUs(hasGPU bool) int {
```
xmonader, line 353: same as above

grid-proxy/tests/queries/node_test.go
```diff
@@ -157,7 +157,7 @@ func TestNode(t *testing.T) {
 		assert.NoError(t, err)
 
 		for _, node := range nodes {
-			assert.Equal(t, node.HasGPU, hasGPU, "has_gpu filter did not work")
+			assert.Equal(t, node.NumGPUs, 1, "has_gpu filter did not work")
```
xmonader: i think the message needs rephrasing, it's not has_gpu anymore

alichaddad: At this point it still is, will change when we settle on the new filters

grid-proxy/internal/explorer/converters.go
alichaddad: the field will be `num_gpus` right? I'd rather keep them consistent

grid-proxy/pkg/types/types.go
alichaddad: I don't think so, since everything in that response is technically a num as well

grid-proxy/internal/explorer/converters.go
xmonader: I agree, let's stay indeed consistent with the field in tfchain, i updated the title to make it `num_gpu` 
https://github.com/threefoldtech/tfchain/issues/756

xmonader (approved):

PR #226, opened 2023-06-26
updates release docs

ashraffouda:
updates release docs to include branching
### Changes

add branching strategy


docs/release.md
```diff
@@ -1,3 +1,9 @@
+# Branching
+
+- Each release is developed on a branch that is derived from the development branch (e.g. development-2.0).
```
xmonader: use undescore in the names instead of dashes

Issue #232, opened 2023-07-03
Grid proxy doesn't filter with the requested keys


Mahmoud-Emad:
## Describe the bug

requested to filter the nodes with the `statue=up` filter key, but got `standby` nodes instead. Also free_cru key doesn't work.

## To Reproduce

- Go to [dev proxy with status up](https://gridproxy.dev.grid.tf/nodes?status=up&ret_count=true&rentable=true&size=10&page=1&free_sru=) or [with free_cru key](https://gridproxy.dev.grid.tf/nodes?status=up&ret_count=true&rentable=true&size=10&page=1&free_sru=&gpu_device=&free_mru=&free_cru=23)
- If there are nodes with a standby status, you can expect to see a screen similar to the one below.

## Screenshots

- ![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/57001890/fac71068-84eb-4a08-9125-09817003851c)
- ![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/57001890/beeed57d-7ef9-4a4f-9242-e784b52f4bdd)

khaledyoussef24:
verified 
can not find any `status :standby` in [proxy while using filter](https://gridproxy.dev.grid.tf/nodes?status=up)
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/15322da7-6c4f-4ad5-a6de-55a17872a091)
and cru filter now works fine also giving the minimum number of cru's
![Screenshot from 2023-07-12 09-20-12](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/0d756a2f-fbf0-460c-8c5f-f0321d45623d)
![Screenshot from 2023-07-12 09-20-01](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/134b920d-d9a3-4f42-b640-75780b282ba3)

PR #235, opened 2023-07-04
enhance node status decision and filter

Omarabdul3ziz:
- edit the node report interval in filtering
- lowercase the key in node.power column while generating
- edit the decideNodeStatus methods
- edit the filter in tests to differentiate between down/standby nodes
- fix for https://github.com/threefoldtech/tfgrid-sdk-go/issues/232

### DIFF
--- a/grid-proxy/internal/explorer/converters.go
+++ b/grid-proxy/internal/explorer/converters.go
@@ -11,16 +11,31 @@ import (
 )
 
 const (
-	nodeUpInterval = -80 * time.Minute
+	// node report its state every 40 mins, if didn't report for 2 cycles, it's marked down
+	// nodes powered by farmerbot report every 24 hours, if didn't report for 1 cycle, it's marked down
+	nodeUpStateFactor         = 2
+	nodeUpReportInterval      = time.Minute * 40
+	nodeStandbyStateFactor    = 1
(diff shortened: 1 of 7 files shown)


grid-proxy/internal/explorer/converters.go
```diff
@@ -11,16 +11,31 @@ import (
 )
 
 const (
-	nodeUpInterval = -80 * time.Minute
+	// node report its state every 40 mins, if didn't report for 2 cycles, it's marked down
+	// nodes powered by farmerbot report every 24 hours, if didn't report for 1 cycle, it's marked down
+	nodeUpStateFactor         = 2
```
alichaddad: These variables are repeated throughout the code, perhaps they should be moved to their own module, and all shared logic can be moved there as well, to make sure that it is consistent across the code base

grid-proxy/internal/explorer/converters.go
```diff
@@ -11,16 +11,31 @@ import (
...
+	nodeUpReportInterval      = time.Minute * 40
+	nodeStandbyStateFactor    = 1
```
alichaddad: Please add docs for this constant

alichaddad (changes requested):

alichaddad (approved):
LGTM

Issue #209, opened 2023-06-20
node gpu info indexer


xmonader:
suggesting to have a background worker that calls listGPUs on all of the nodes and insert them in the db, full gpus info per node maybe should be available on `/nodes/<nodeid>/gpus`

xmonader:
This also means we can expand the filters by gpu device type as well by searching for devices too along with `has_gpu` feature

alichaddad:
The indexer will run periodically and query all the available nodes, it will then retrieve the GPU information for each node if available and store it in a new table, the fields of that table would be:

- `id` gpu id
- `node_id` the id of the node
- `vendor` GPU vendor name
- `device` GPU device name
- `contract`


While we could store the device and vendor id as well for simplifying querying, it seems a bit redundant and we can still use the id to query via device/vendor id if needed.

Listing nodes now will allow filtering based on device and gpu partial names and if there is a need to list to list available GPU vendors/devices a new endpoint can be added to retrieve that information from the new `node_gpu` table.

muhamadazmy:
I think it's important to also store the Vendor ID and the Device ID (numeric values) that are part of the ID. Later to search for a specific device, the user can search the full PCI database for a specific device name, then we use that (numeric) ID to search the database.

> IMHO the device and vendor names are the redundant data not the other way around

If you also going to store the device and vendor names, it's important to also keep the numeric ids. 

I would also drop the contract from the database, since this is useless during the search and is useful only to someone who has the node rented, hence can then get in real time from the node directly. the contract can be out of date as well.

A free node for renting can only has all contract ids values as zeros.

muhamadazmy:
As @alichaddad brought my attention that the device/vendor ids are already in the ID of the device we can then not store the numeric values but then we can do partial string matches on the ID to match on specific vendor and or device with their ids

Mahmoud-Emad:
Hey @alichaddad 
What is the status here?

alichaddad:
Code is ready for review should be done maximum tomorrow

alichaddad:
PR review is still in progress

A-Harby:
Verified,

Proxy calls are added and works successfully, and docs are updated.
https://gridproxy.dev.grid.tf/nodes?gpu_vendor_name=Advanced%20Micro%20Devices
https://gridproxy.dev.grid.tf/nodes?gpu_device_name=Navi%2031
https://gridproxy.dev.grid.tf/nodes?gpu_available=true
[...]

PR #236, opened 2023-07-04
add gpu indexer

alichaddad:
### Description

Introduce GPU indexer to store node GPU information

### Changes

- Implement the indexer that runs periodically and fetch GPU information and stores them in new table

### Related Issues

[List of related issues](https://github.com/threefoldtech/tfgrid-sdk-go/issues/209)

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


grid-proxy/cmds/proxy_server/main.go
```diff
@@ -96,12 +101,24 @@ func main() {
...
 		log.Fatal().Err(err).Msg("failed to create relay client")
 	}
 
-	s, err := createServer(f, GitCommit, relayClient)
+	db, err := db.NewPostgresDatabase(f.postgresHost, f.postgresPort, f.postgresUser, f.postgresPassword, f.postgresDB)
+	if err != nil {
+		log.Fatal().Err(err).Msg("couldn't get postgres client")
+	}
+
+	worker, err := gpuindexer.NewNodeGPUIndexer(ctx, f.relayURL, f.mnemonics, sub, db, f.indexerCheckIntervalMins, f.workerBatchSize)
```
xmonader: This is a little bit misleading, It should return the concrete Indexer and on the indexer you do indexer.StartWorkers .. etc

grid-proxy/internal/gpuindexer/gpuindexer.go
```diff
@@ -0,0 +1,176 @@
...
+		n.nodesGPUChan <- nodesGPU
+	}
+
+}
+
+func (n *NodeGPUIndexer) Start(ctx context.Context) {
+	ticker := time.NewTicker(n.checkInterval)
+	log.Info().Msg("GPU indexer started")
+	go n.worker(ctx)
+	n.run(ctx)
```
ashraffouda: looks good, but isn't it more clean if we did it like 
```
go n.worker(ctx)
go n.run(ctx) and inside run we do the ticker?

**Review by ashraffouda — CHANGES_REQUESTED** [review]

File: grid-proxy/internal/gpuindexer/gpuindexer.go [hunk 3]
```diff
@@ -0,0 +1,176 @@
...
+	rmbTypes "github.com/threefoldtech/tfgrid-sdk-go/rmb-sdk-go/direct/types"
+)
+
+const workerCleanupInterval = 10 * time.Second
+
+type NodeGPUIndexer struct {
+	db              db.Database
+	relayClient     *direct.Client
+	nodesGPUChan    chan []types.NodeGPU
+	checkInterval   time.Duration
```
xmonader: Is that indexing interval?

grid-proxy/internal/gpuindexer/gpuindexer.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/rs/zerolog/log"
+	substrate "github.com/threefoldtech/tfchain/clients/tfchain-client-go"
+	"github.com/threefoldtech/tfgrid-sdk-go/grid-proxy/internal/explorer/db"
+	"github.com/threefoldtech/tfgrid-sdk-go/grid-proxy/pkg/types"
+	"github.com/threefoldtech/tfgrid-sdk-go/rmb-sdk-go"
+	"github.com/threefoldtech/tfgrid-sdk-go/rmb-sdk-go/direct"
+	rmbTypes "github.com/threefoldtech/tfgrid-sdk-go/rmb-sdk-go/direct/types"
+)
+
+const workerCleanupInterval = 10 * time.Second
```
xmonader: Does this mean the worker dies after 10 seconds? or that's the wait time before bucketing the work units?

grid-proxy/cmds/proxy_server/main.go
alichaddad: The name of the variable is misleading, this is the indexer, will change the name

grid-proxy/internal/gpuindexer/gpuindexer.go
alichaddad: This is done this way so that `run` runs at the start of the code, otherwise it will have to wait for the interval to pass before starting the indexer job.

grid-proxy/internal/gpuindexer/gpuindexer.go
alichaddad: Yes

grid-proxy/internal/gpuindexer/gpuindexer.go
alichaddad: That's the buffer timeout, the interval in which the buffer will be emptied

grid-proxy/internal/gpuindexer/gpuindexer.go
```diff
@@ -0,0 +1,176 @@
...
+		select {
+		case <-ticker.C:
+			n.run(ctx)
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func (n *NodeGPUIndexer) run(ctx context.Context) {
```
xmonader: name should be queryGridNodes

grid-proxy/internal/gpuindexer/gpuindexer.go
ashraffouda: > This is done this way so that `run` runs at the start of the code, otherwise it will have to wait for the interval to pass before starting the indexer job.

yeah. np for that I mean this to be done inside the run method

grid-proxy/internal/gpuindexer/gpuindexer.go
```diff
@@ -0,0 +1,176 @@
...
+		nodesGPU[i].NodeTwinID = response.Source.Twin
+	}
+	// Will be using only one worker giving the current amount of nodes supporting GPUs
+	if len(nodesGPU) != 0 {
+		n.nodesGPUChan <- nodesGPU
+	}
+
+}
+
+func (n *NodeGPUIndexer) Start(ctx context.Context) {
```
xmonader: I'd do the following here

```
go queryGridNodes()
go gpuNodeResultsBatcher()
go gpuBatchesDBUpserter()
```



queryGridNodes (called `run`), this is being ran to index the gridNodes maybe every 1hr? could be less, I'm thinking also of spliting worker to reflect the behavior into two (gpuNodeResultsBatcher, gpuBatchesDBUpserter) where we separate the concerns of preparing batches and inserting them into the database 

I probably mislead you on Line:120 `// Will be using only one worker giving the current amount of nodes supporting GPUs` as we should never assume such, we should assume the other way around, the grid will have e.g 4000 nodes with GPUs

So, There's a goroutine that queries the nodes regularly, hands over the results of each node to the nodesGPUChan, then there's one or more workers creating batches, and there's one or more workers upserting the batches into the database


I would also keep all of these workers alive for as running as the app, so probably will need to have the ctx passed around to them

grid-proxy/internal/gpuindexer/gpuindexer.go
```diff
@@ -0,0 +1,176 @@
...
+	errResp := response.GetError()
+
+	if errResp != nil {
+		log.Error().Msg(errResp.Message)
+		return
+	}
+
+	resp := response.GetResponse()
+	if resp == nil {
+		log.Error().Msg("received a non response envelope")
```
xmonader, line 179: I'd be curious to see the resp, please attach it to the error message

grid-proxy/internal/gpuindexer/gpuindexer.go
```diff
@@ -0,0 +1,176 @@
...
+		return
+	}
+
+	if response.Schema == nil || *response.Schema != rmb.DefaultSchema {
+		log.Error().Msgf("invalid schema received expected '%s'", rmb.DefaultSchema)
```
xmonader, line 184: can we please attach the received one as well

grid-proxy/internal/gpuindexer/gpuindexer.go
```diff
@@ -0,0 +1,176 @@
...
+		return
+	}
+
+	output := response.Payload.(*rmbTypes.Envelope_Plain).Plain
+
+	var nodesGPU []types.NodeGPU
+	err := json.Unmarshal(output, &nodesGPU)
+	if err != nil {
+		log.Error().Err(err).Msg("failed to unmarshal GPU information response")
```
xmonader: can we attach the bytes here please

ashraffouda (approved):

grid-proxy/cmds/proxy_server/main.go
```diff
@@ -96,12 +105,33 @@ func main() {
...
 		log.Fatal().Err(err).Msg("failed to create relay client")
 	}
 
-	s, err := createServer(f, GitCommit, relayClient)
+	db, err := db.NewPostgresDatabase(f.postgresHost, f.postgresPort, f.postgresUser, f.postgresPassword, f.postgresDB)
+	if err != nil {
+		log.Fatal().Err(err).Msg("couldn't get postgres client")
+	}
+
+	worker, err := gpuindexer.NewNodeGPUIndexer(
```
xmonader: can we rename worker -> indexer?

grid-proxy/internal/gpuindexer/gpuindexer.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+
+		limit.Page++
+	}
+}
+
+func (n *NodeGPUIndexer) gpuBatchesDBUpserter(ctx context.Context) {
+	for {
+		select {
+		case gpuBtach := <-n.nodesGPUBuBatchesChan:
```
AbdelrahmanElawady: I think it's `gpuBatch`

grid-proxy/tests/queries/local_client.go
```diff
@@ -398,6 +398,41 @@ func (g *GridProxyClientimpl) Counters(filter proxytypes.StatsFilter) (res proxy
 	return
 }
 
+func gpuSatisfies(data *DBData, node node, f proxytypes.NodeFilter) bool {
+	gpu := data.gpus[node.twin_id]
+
+	if f.GpuDeviceName != nil {
+		if !strings.Contains(strings.ToLower(gpu.device), *f.GpuDeviceName) {
```
AbdelrahmanElawady, line 405: I think `contains` here is not the best way for GPU device name as some GPU names get a suffix `ti`, `super` or `xtx` so there might be devices that get reported inaccurately.

for example: 4090 and 4090ti, 7900xt and 7900xtx

or is that intended behavior to report all that start with the needed GPU name?

grid-proxy/internal/gpuindexer/gpuindexer.go
```diff
@@ -0,0 +1,205 @@
...
+		go n.gpuNodeResultsBatcher(ctx)
+	}
+
+	for i := 0; i < n.nodesGPUBufferWorkers; i++ {
+		go n.gpuBatchesDBUpserter(ctx)
+	}
+
+	go n.queryGridNodes(ctx)
+	go n.gpuNodeResultsBatcher(ctx)
+	go n.gpuBatchesDBUpserter(ctx)
```
AbdelrahmanElawady: calls to these functions are repeated in the for loop and outside, is there a reason for that?

grid-proxy/tests/queries/local_client.go
alichaddad: This is only done for a specific test case with known input, the tests are in the middle of a refactor right now so this isn't a problem.

AbdelrahmanElawady (approved):

Issue #249, opened 2023-07-20
Swagger docs don't cover all node filter keys 


Mahmoud-Emad:
## Describe the bug

The swagger docs don't support the filter by total values or `sru, cru, mru, hru` and it's supported to filter by them as seen [here](https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/grid-proxy/internal/explorer/helpers.go#L132-L135) 

## To Reproduce
- Go to the [swagger API](https://gridproxy.dev.grid.tf/swagger/index.html#/GridProxy/get_nodes)

mohamedamer453:
total_cru should be in `vCores` not in `bytes`

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/101194226/2abf6b34-c990-48fc-bbed-71cfae1b55c9)

xmonader:
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/64129/543575b4-2817-4fd5-bedc-ae08cd985b1c)
https://gridproxy.qa.grid.tf/swagger/index.html#/GridProxy/get_nodes

PR #251, opened 2023-07-20
remove dependency on has_gu by using node_gpu table

alichaddad:
### Description

Step depending on `has_gpu` and use `node_gpu` table instead.

### Changes

- Calculate `gpu_num` based on the count in `node_gpu`
- Filter `has_gpu` using `node_gpu`
- Add missing API docs

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/249

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation


grid-proxy/tests/queries/local_client.go
```diff
@@ -283,6 +283,10 @@ func (g *GridProxyClientimpl) Twins(filter proxytypes.TwinFilter, limit proxytyp
 }
 func (g *GridProxyClientimpl) Node(nodeID uint32) (res proxytypes.NodeWithNestedCapacity, err error) {
 	node := g.data.nodes[uint64(nodeID)]
+	numGPU := 0
+	if _, ok := g.data.gpus[node.twin_id]; ok {
+		numGPU = 1
```
ashraffouda, line 288: why is this always one? shouldn't we get how many gpus attached to this node? should we make use of the returned value instead of hardcoding to 1?

ashraffouda (approved):

Issue #263, opened 2023-08-01
support some queries on farms endpoint


AhmedHanafy725:
required for https://github.com/threefoldtech/tfgrid-sdk-ts/issues/491

- [ ] node status
- [ ] node rented by (twin id)
- [ ] node certified
- [ ] node has gpu
- [ ] has nodes available for (twin id)

xmonader:
@MarioBassem please create a ticket to upgrade the proxy

mohamedamer453:
Verified on devnet.

All of the mentioned queries were added and documented in the swagger page.

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/101194226/7bb58197-a907-4d79-ad52-b535e6a6575c)

### Node status

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/101194226/92b59128-49a9-4ca5-b9c1-239f600ef6cb)

### Node rented by

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/101194226/c35c35d4-49cd-4489-9076-538084df2672)

### Nodes certified

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/101194226/420dfada-092b-4b3c-bb63-6cebbf1a0f16)

### Node has gpu

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/101194226/ab93d318-3f90-4621-a2cf-cf1cc83e5bba)

### Node available for

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/101194226/8eecf5fa-c19f-4aee-a2e6-aaf75234d544)

PR #266, opened 2023-08-01
update farm filter

mariobassem:
add more fields to farm filter

### Description

This PR adds NodeStatus, NodeRentedBy, NodeAvailableFor, NodeHasGPU, NodeCertified fields for farm filter

### Related Issues

- #263

### DIFF
--- a/grid-proxy/docs/docs.go
+++ b/grid-proxy/docs/docs.go
@@ -247,6 +247,36 @@ const docTemplate = `{
                         "description": "Min free reservable sru for at least a single node that belongs to the farm, in bytes",
                         "name": "node_free_sru",
                         "in": "query"
+                    },
+                    {
+                        "type": "string",
+                        "description": "Node status for at least a single node that belongs to the farm",
+                        "name": "node_status",
+                        "in": "query"
+                    },
+                    {
+                        "type": "integer",
+                        "description": "Twin ID of user who has at least one rented node in the farm",
+                        "name": "node_rented_by",
+                        "in": "query"
+                    },
+                    {
+                        "type": "integer",
+                        "description": "Twin ID of user for whom there is at least one node that is available to be deployed to in the farm",
+                        "name": "node_available_for",
+                        "in": "query"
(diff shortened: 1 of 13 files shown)


Omarabdul3ziz:
Good work from a good developer ya Mario :D

Just a couple of small comments:
- Can we update the swagger docs with the new farm filters?

- The `serializeFilters` function helps reduce code duplication, but it's only applied to farm/node filters. can we apply it to twin, contract, and stats filters too?


- I am not sure if the certification for the farm and nodes on it should be the same. but shouldn't `NodeCertified` filter check on the certification on the node table not in farm?
  https://github.com/threefoldtech/tfgrid-sdk-go/blob/4965ea56411dd7b02a7219c63f64a51278f00daf/grid-proxy/internal/explorer/db/postgres.go#L623-L625
  tried to filter the certified node with `/nodes?certification_type=certified` and i got 3 node in farm 1 so `/farms?node_certified=true` should list it but it returns empty list. farm 1 itself has `"certificationType": "NotCertified"`


- In `/nodes?available_for=89` filter the list is ordered to return the rented nodes at the begging can we do that with `/farms?node_available_for=89` too
  https://github.com/threefoldtech/tfgrid-sdk-go/blob/4965ea56411dd7b02a7219c63f64a51278f00daf/grid-proxy/internal/explorer/db/postgres.go#L561-L563


- Things look good but i got these error on logs each time `/farms` endpoint is requested

mariobassem:
fixed above comments except for the `SerializeFilter` part, it's done in another PR #244. I only added it here because I needed it while running farm tests.

Omarabdul3ziz (approved):
LGTM

alichaddad (approved):

mariobassem:
had a small discussion with @alichaddad about the farms query. he had a concern about what would happen if a node had multiple gpus. we reached a conclusion that whatever the node data that resulted, the query result would be grouped by farm information in the end, and all node information would be discarded.

Issue #306, opened 2023-08-08
Some unit records couldn't parsed using ParseUint


sameh-farouk:
## Describe the bug

From Gridproxy logs on dev network

```bash
error couldn't parse total_sru strconv.ParseUint: parsing "512067475865.6": invalid syntax: bad request
error couldn't parse total_cru strconv.ParseUint: parsing "44.54": invalid syntax: bad request
```

Omarabdul3ziz:
the issue happened because of a bad request from the client. we are handling all the numeric value filters as `uint64` so passing a float value to `total_sru`, `total_cru`, or ... will raise an error. 

but anyway we are heading to use the `scheme` package to handle all the query params. linked the issue to the scheme pr


this part is removed in the mentioned pr
https://github.com/threefoldtech/tfgrid-sdk-go/blob/cca933e1bf26694bff86e85da61bfe973ad3c0be/grid-proxy/internal/explorer/helpers.go#L96-L99

Issue #269, opened 2023-08-02
use schema package to encode and decode query paramters


mariobassem:
currently, http query parameters are encoded and decoded using local converters, which may be error prone if some type changes happen. luckily there is a package that does this already [schema](https://github.com/gorilla/schema). 

i think it would be better to use it and not bother with the conversion again.

xmonader:
It's definitely a step in the right direction.

mariobassem:
rebased the pr branch on latest development

Omarabdul3ziz:
track updates on the pr discussion

PR #315, opened 2023-08-10
use schema package

mariobassem:
### Description

This pr uses [schema](https://github.com/gorilla/schema) package to encode and decode query parameters instead of using our own converters.

### Related Issues

- #269 
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/306


grid-proxy/internal/explorer/db/postgres.go
```diff
@@ -762,7 +762,7 @@ func (d *PostgresDatabase) GetContracts(filter types.ContractFilter, limit types
 			`LEFT JOIN (
 				SELECT 
 					contract_bill_report.contract_id,
-					COALESCE(json_agg(json_build_object('amountBilled', amount_billed, 'discountReceived', discount_received, 'timestamp', timestamp)), '[]') as billings
+					COALESCE(json_agg(json_build_object('amount_billed', amount_billed, 'discount_received', discount_received, 'timestamp', timestamp)), '[]') as billings
```
rawdaGastan: won't this change create any bugs ?

grid-proxy/internal/explorer/param_parser.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/pkg/errors"
+)
+
+func parseQueryParams(r *http.Request, values ...interface{}) error {
+	params := r.URL.Query()
+
+	decoder := schema.NewDecoder()
+	decoder.IgnoreUnknownKeys(true)
+
+	for idx := range values {
```
rawdaGastan: you can use `for _, value := range values` directly

grid-proxy/internal/explorer/param_parser.go
```diff
@@ -0,0 +1,23 @@
...
+		if err := decoder.Decode(values[idx], params); err != nil {
+			return errors.Wrap(err, "failed to decode filter parameters")
```
rawdaGastan: add the failed value to the error

grid-proxy/internal/explorer/param_parser_test.go
```diff
@@ -0,0 +1,75 @@
...
+		types.Limit{},
+	}
+
+	for _, filter := range filters {
+		testDecode(t, filter)
+	}
+}
+
+func testDecode(t *testing.T, filter interface{}) {
+	req, err := http.NewRequest(http.MethodGet, "gridproxy.com", nil)
```
rawdaGastan, line 32: what is `gridproxy.com` ?

grid-proxy/internal/explorer/param_parser_test.go
```diff
@@ -0,0 +1,75 @@
...
+	assert.NoError(t, err)
+
+	want := reflect.New(reflect.TypeOf(filter)).Elem()
+
+	urlValues := url.Values{}
```
rawdaGastan, line 37: would be `var urlValues type`

grid-proxy/internal/explorer/param_parser_test.go
```diff
@@ -0,0 +1,75 @@
...
+	for i := 0; i < want.NumField(); i++ {
+		var x, y reflect.Value
+		if want.Field(i).Type().Kind() == reflect.Pointer {
+			x = reflect.New(want.Field(i).Type().Elem())
+			y = x.Elem()
+		} else {
+			x = reflect.New(want.Field(i).Type()).Elem()
+			y = x
```
rawdaGastan: y can be `x.Elem()` in both cases

```
if want.Field(i).Type().Kind() == reflect.Pointer {
	x = reflect.New(want.Field(i).Type().Elem())
} else {
	x = reflect.New(want.Field(i).Type())
}
y = x.Elem()
```

grid-proxy/internal/explorer/param_parser_test.go
```diff
@@ -0,0 +1,75 @@
...
+		want.Field(i).Set(x)
+
+		schemaTag, ok := want.Type().Field(i).Tag.Lookup("schema")
+		if !ok {
+			t.Fatalf("schema tag must be provided for all fields in a filter")
+		}
+
+		name := strings.Split(schemaTag, ",")[0]
+		if want.Field(i).Type().Kind() == reflect.Slice {
+			for j := 0; j < want.Field(i).Len(); j++ {
```
rawdaGastan, line 56: cannot it be a for range instead?

grid-proxy/internal/explorer/server.go
```diff
@@ -59,10 +59,12 @@ const (
 // @Failure 500 {object} string
 // @Router /farms [get]
 func (a *App) listFarms(r *http.Request) (interface{}, mw.Response) {
-	filter, limit, err := a.handleFarmRequestsQueryParams(r)
-	if err != nil {
+	filter := types.FarmFilter{}
+	limit := types.DefaultLimit()
```
rawdaGastan, line 60: why using default one? not `types.Limit`

grid-proxy/pkg/client/grid_client.go
```diff
@@ -5,12 +5,16 @@ import (
...
 	"net/http"
+	"net/url"
 	"strconv"
 
+	"github.com/gorilla/schema"
 	"github.com/pkg/errors"
 	"github.com/threefoldtech/tfgrid-sdk-go/grid-proxy/pkg/types"
 )
 
+var encoder = schema.NewEncoder()
```
rawdaGastan, line 19: why it's defined here not in the function?

grid-proxy/pkg/client/grid_client.go
```diff
@@ -219,17 +220,55 @@ func (g *Clientimpl) NodeStatus(nodeID uint32) (res types.NodeStatus, err error)
...
+	if err := json.NewDecoder(response.Body).Decode(&res); err != nil {
 		return res, err
 	}
 	return
 }
+
+func (g *Clientimpl) makeRequest(path string, params ...interface{}) (*http.Response, error) {
+	url, err := g.prepareURL(path, params...)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to prepare url")
```
rawdaGastan: encode the `url` in the error
please apply for others

grid-proxy/pkg/client/grid_client.go
```diff
@@ -219,17 +220,55 @@ func (g *Clientimpl) NodeStatus(nodeID uint32) (res types.NodeStatus, err error)
...
+
+	baseURL := g.endpoint
+	resource := fmt.Sprintf("/%s", path)
+
+	u, err := url.ParseRequestURI(baseURL)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to parse request URI")
+	}
+
+	u.Path = resource
```
rawdaGastan: couldn't it be `fmt.Sprintf("/%s", path)` directly?

grid-proxy/pkg/client/grid_client.go
```diff
@@ -219,17 +220,55 @@ func (g *Clientimpl) NodeStatus(nodeID uint32) (res types.NodeStatus, err error)
...
+			return "", errors.Wrap(err, "failed to encode query params")
+		}
+	}
+
+	baseURL := g.endpoint
+	resource := fmt.Sprintf("/%s", path)
+
+	u, err := url.ParseRequestURI(baseURL)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to parse request URI")
```
rawdaGastan: encode `baseUrl` in it

grid-proxy/pkg/client/grid_client.go
```diff
@@ -219,17 +220,55 @@ func (g *Clientimpl) NodeStatus(nodeID uint32) (res types.NodeStatus, err error)
 
 // Counters return statistics about the grid
 func (g *Clientimpl) Counters(filter types.StatsFilter) (res types.Counters, err error) {
-	query := statsParams(filter)
-	req, err := http.Get(g.url("stats%s", query))
+	response, err := g.makeRequest("stats", filter)
 	if err != nil {
-		return
+		return types.Counters{}, err
```
rawdaGastan: why not just a `return`?

grid-proxy/pkg/client/grid_client.go
```diff
@@ -219,17 +220,55 @@ func (g *Clientimpl) NodeStatus(nodeID uint32) (res types.NodeStatus, err error)
...
 	}
-	if req.StatusCode != http.StatusOK {
-		err = parseError(req.Body)
-		return
-	}
-	if err := json.NewDecoder(req.Body).Decode(&res); err != nil {
+
+	if err := json.NewDecoder(response.Body).Decode(&res); err != nil {
```
rawdaGastan: I think no need to check error we can return directly as it's the end of the function

grid-proxy/pkg/client/grid_client.go
```diff
@@ -72,114 +76,108 @@ func (g *Clientimpl) Ping() error {
...
-	query := nodeParams(filter, limit)
-	req, err := http.Get(g.url("nodes%s", query))
+	response, err := g.makeRequest("nodes", filter, limit)
 	if err != nil {
-		return
-	}
-	if req.StatusCode != http.StatusOK {
-		err = parseError(req.Body)
-		return
+		return nil, 0, err
```
rawdaGastan: I think we can just `return` here
please check others

grid-proxy/pkg/client/grid_client.go
```diff
@@ -72,114 +76,108 @@ func (g *Clientimpl) Ping() error {
...
 	}
-	data, err := io.ReadAll(req.Body)
+
+	data, err := io.ReadAll(response.Body)
 	if err != nil {
 		return
 	}
+
 	err = json.Unmarshal(data, &res)
```
rawdaGastan: `if err = json.Unmarshal(data, &res); err != nil` would be good as you used before
please apply for others

grid-proxy/pkg/client/grid_client.go
```diff
@@ -72,114 +76,108 @@ func (g *Clientimpl) Ping() error {
...
+	data, err := io.ReadAll(response.Body)
 	if err != nil {
 		return
 	}
 	err = json.Unmarshal(data, &res)
 	if err != nil {
 		return
 	}
+
 	for idx := range res {
```
rawdaGastan: use the value instead of the index

grid-proxy/pkg/client/grid_client.go
```diff
@@ -72,114 +76,108 @@ func (g *Clientimpl) Ping() error {
...
 		if res[idx].Type == "node" {
 			res[idx].Details = types.NodeContractDetails{
-				NodeID:            uint(res[idx].Details.(map[string]interface{})["nodeId"].(float64)),
+				NodeID:            uint(res[idx].Details.(map[string]interface{})["node_id"].(float64)),
```
rawdaGastan, line 164: is it ok as a change?

grid-proxy/pkg/client/grid_client_test.go
```diff
@@ -28,6 +30,67 @@ var (
 	NodeStatusExample = MarshalNodeStatus([]byte(NodeStatusExampleStr))
 )
 
+func nodesFilterValues() (types.NodeFilter, types.Limit, string) {
+	Up := "up"
```
rawdaGastan, line 35: suggestion:
we can create a function with any value to be used directly in the filter instead of defining all the values:
```go
func ref[T any](v T) *T {
	return &v
}
```

grid-proxy/pkg/client/grid_client_test.go
```diff
@@ -28,6 +30,67 @@ var (
...
+		Domain:       &trueVal,
+		Rentable:     &falseVal,
+		RentedBy:     &ints[5],
+		AvailableFor: &ints[6],
+	}
+	l := types.Limit{
+		Page: 12,
+		Size: 13,
+	}
+	return f, l, "status=up&free_mru=1&free_hru=2&free_sru=3&country=Egypt&city=Mansoura&farm_name=Freefarm&farm_ids=1&farm_ids=2&free_ips=4&ipv4=true&ipv6=false&domain=true&rentable=false&rented_by=5&available_for=6&page=12&size=13&ret_count=false&randomize=false"
```
rawdaGastan, line 63: shouldn't we encode instead of writing the string?

grid-proxy/pkg/client/grid_client_test.go
```diff
@@ -270,3 +341,62 @@ func testSuccess(t *testing.T, f ProxyFunc) {
...
+	dedicated := true
+	filter := types.NodeFilter{
+		Status:    &status,
+		FreeMRU:   &freeMRU,
+		FarmIDs:   farmIDs,
+		Dedicated: &dedicated,
+	}
+	limit := types.DefaultLimit()
+
+	endpoint := "http://www.gridproxy.com"
```
rawdaGastan, line 359: I don't think we should use this url? Please, check other urls

grid-proxy/pkg/client/grid_client_test.go
```diff
@@ -270,3 +341,62 @@ func testSuccess(t *testing.T, f ProxyFunc) {
...
+
+	gotURL, err := url.Parse(got)
+	assert.NoError(t, err)
+
+	assert.Equal(t, wantURL.Path, gotURL.Path)
+	assert.True(t, reflect.DeepEqual(wantURL.Query(), gotURL.Query()))
+}
+
+func TestEncodeEmptyNodeFilter(t *testing.T) {
+	values := url.Values{}
```
rawdaGastan: `var values url.Values`? please apply for others

grid-proxy/pkg/types/contracts.go
```diff
@@ -0,0 +1,50 @@
+package types
+
+type ContractDetails interface {
```
rawdaGastan, line 6: add comments for exported types.
check others

grid-proxy/pkg/types/counters.go
```diff
@@ -0,0 +1,49 @@
...
+	NodesDistribution map[string]int64 `json:"nodes_distribution" gorm:"-:all"`
+	GPUs              int64            `json:"gpus"`
+}
+
+// StatsFilter statistics filters
+type StatsFilter struct {
+	Status *string `schema:"status,omitempty"`
+}
+
+type NodeStatisticsResources struct {
```
rawdaGastan, line 28: add comments for exported types.
check all please

grid-proxy/pkg/types/farms.go
```diff
@@ -0,0 +1,44 @@
+package types
+
+type Farm struct {
```
rawdaGastan, line 4: add comments for exported types.
check all please

grid-proxy/pkg/types/limit.go
```diff
@@ -0,0 +1,18 @@
...
+
+// Limit used for pagination
+type Limit struct {
+	Size      uint64 `schema:"size"`
+	Page      uint64 `schema:"page"`
+	RetCount  bool   `schema:"ret_count"`
+	Randomize bool   `schema:"randomize"`
+}
+
+func DefaultLimit() Limit {
```
rawdaGastan, line 12: add comments for exported func.

grid-proxy/pkg/types/nodes.go
```diff
@@ -0,0 +1,139 @@
+package types
+
+import "github.com/threefoldtech/zos/pkg/gridtypes"
+
+type Location struct {
```
rawdaGastan, line 6: add comments for exported types.
check all please

grid-proxy/pkg/types/twins.go
```diff
@@ -0,0 +1,16 @@
+package types
+
+type Twin struct {
```
rawdaGastan, line 4: add comments for exported type

grid-proxy/pkg/types/version.go
```diff
@@ -0,0 +1,5 @@
+package types
+
+type Version struct {
```
rawdaGastan, line 4: add comments for exported type

rawdaGastan (changes requested):

rawdaGastan:
Thanks for cleaning up! 💯

grid-proxy/internal/explorer/param_parser_test.go
Omarabdul3ziz: I think it is just any domain to test decoding the params into, no actual request is done

Omarabdul3ziz:
resolved the conflict, but noticed some json tags changes could break the clients, will revert that.
drafting the pr till work is done.

grid-proxy/internal/explorer/server.go
Omarabdul3ziz: The old code used `getLimit` method to set the size=50 and page=1 if not specified. 
The  `DefaultLimit` ensures that there are values for these fields, even if they are missing in the request before the parseQueryParam

grid-proxy/pkg/client/grid_client_test.go
Omarabdul3ziz: from what I gather, he hard-codes the expected values here from the defined f and l
then in the tests, can assert the equality of this string value to the ones from decoding both (f, l)

grid-proxy/pkg/client/grid_client.go
```diff
@@ -365,3 +435,25 @@ func (g *Clientimpl) newHTTPClient() *http.Client {
 		},
 	}
 }
+
+func (g *Clientimpl) prepareURL(path string, params ...interface{}) (string, error) {
+	values := url.Values{}
```
rawdaGastan, line 440: var values url.Values? please apply for others

rawdaGastan:
some simple comments aren't resolved. please check them :D

grid-proxy/pkg/client/grid_client.go
Omarabdul3ziz: it needs to be initialized

grid-proxy/pkg/client/grid_client_test.go
Omarabdul3ziz: it needs to be initialized

grid-proxy/pkg/client/grid_client.go
Omarabdul3ziz: declared as a global variable. it is being used in the client_test methods

grid-proxy/internal/explorer/param_parser_test.go
Omarabdul3ziz: looks like we can't range over `want.Field(i)`

grid-proxy/internal/explorer/db/postgres.go
```diff
@@ -704,7 +704,7 @@ func (d *PostgresDatabase) GetFarms(ctx context.Context, filter types.FarmFilter
 		q = q.Where("public_ip_count.total_ips >= ?", *filter.TotalIPs)
 	}
 	if filter.StellarAddress != nil {
-		q = q.Where("farm.stellar_address = ?", *filter.StellarAddress)
+		q = q.Where("COALESCE(farm.stellar_address, '') = ?", *filter.StellarAddress)
```
muhamadazmy: I don't get the overuse of COALESCE, why do u need to use it here? if you already compaing to a given value it won't mater if its `nil = ?` or ` '' = ?` right? so you probably don't need to use COALESCE unless you have an explansion that i don't know why it's needed

rawdaGastan (approved):

grid-proxy/pkg/client/grid_client.go
muhamadazmy: In `go` declaring a variable of a **struct** already initalize it to zero value, so 
```go
type Something struct{
   // data
}
var obj Something
// and
obj := Something{}
// are exactly the same
```

But this doesn't apply for data structures that are not struct, like maps, channels, and arrays, So it's only correct to do `url.values` since Values is actually a map
so the only way possible is
```go
values := url.Values{}
// OR 
values := make(url.Values)
```

muhamadazmy (approved):

grid-proxy/internal/explorer/db/postgres.go
Omarabdul3ziz: removed.
while removing that,  discovered another problem, the current implementation of parsing the params doesn't ignore the empty string values so i updated the code to remove any empty values in the values map before the decoding

muhamadazmy (approved):

Issue #308, opened 2023-08-08
improve grid proxy queries 


xmonader:
proper indices and queries, make sure to include performance changes

xmonader:
What is the status here? it's linked to a closed PR?

if already addressed in another PR, please make sure to link it here and update the project to 3.12

mariobassem:
the pr tried to introduce a way to improve contracts endpoint performance, but it was closed in favor of #356 

this is already addressed in some issues:
- #347
- #265 
- #418 
-  there should be an issue to improve farms endpoint performance

i did a pr #419 that did a few things:
- database schema changes
- added basic `id` indices (farm_id, node_id, contract_id, ...)
- update/improvement on farms query

Omarabdul3ziz:
WIP:
reviewed the mentioned pr and was working on separating the schema change part in another branch cause it may need updates in graphql and data migrationsm or mario was suggesting to create the aggregated tables on setup query.

xmonader:
Status?

Omarabdul3ziz:
there is no progress here yet, there were higher-priority issues I was working on.
I started investigating the options we can do to optimize the queries and will keep the issue updated

Omarabdul3ziz:
update (wip):
- created indices on the IDs for the (node/farm/contract/twin) tables, the most used column with where/join clauses.
- explored the possibility of combining `Count` and `Scan` queries 
  it seems that each request happens twice one time to get the count and the other to get the rows itself. tried to minimize this by combining the two calls in one but there were some issues:
  - broked the paginations params `page` and `size` since the `limit`/`offset` will conflict with the `count` clause.
  - anyway `gorm` will handle this `q.Count(&count).Limit(size).Offset((page - 1) * size).Find(&nodes).Error` as two calls and will not introduce any enhancements so things are as is.
    ![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/32527347/b92cc319-036a-43f7-a5ec-9386de6069d2)
- started experimenting with adding some `materialized views` which will precompute and store the results of some of the complex queries like `farmTableQuery`/`nodeTableQuery` and then it will be faster to filter, this will require a refresher every now and then so we may add 
  - a triggered refresher by the effected columns in this view
  - a scheduled refresher that repeatedly works on interval

Omarabdul3ziz:
update (wip):
[...]

Issue #231, opened 2023-07-03
investigate other ways to perform contracts query


mariobassem:
currently, the sql query used in fetching contracts from the database contains a subquery to retrieve relevant information, which makes performance a bit bad, as the sql command alone could take ~100ms.

Omarabdul3ziz:
update:
merged now but waiting for a release to move to verification

A-Harby:
Verified, Devnet.
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/00f03f52-f84b-4d6a-aecb-78c473a68aec)


Docs was udpated, https://github.com/threefoldtech/tfgrid-sdk-go/blob/ac6592e8a210a4a98b1349909a331ee10c71b6d9/grid-proxy/docs/docs.go#L135C38-L135C38 .

As well as Swagger, https://gridproxy.dev.grid.tf/swagger/index.html#/Contract/get_contracts__contract_id_.

PR #317, opened 2023-08-10
create redundant table for aggregated bill reports

mariobassem:
### Description

this pr creates a redundant table for aggregated bill reports, and lets the contracts sql query make use of it.

### Related Issues

- #231 
- #308


mariobassem:
it is worth noting that currently some contracts on devnet have +30K bill reports, so even with the redundant table, it takes time just to fetch them.

another solution is to separate bill reports information from contracts.

mariobassem:
will go with another solution in #356

Issue #329, opened 2023-08-15
Add country filter on farms endpoint


AhmedHanafy725:
The playground is using now the farm filters at the beginning so the country should be there to provide the farms within a certain country.

Related: https://github.com/threefoldtech/tfgrid-sdk-ts/issues/862

A-Harby:
Verified Devnet.
Docs and swagger was updated with country filter.
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/4c4155a7-7d43-4fc2-a2d7-b84c01742c31)
And Endpoint is working fine with both sensitive and insensitive cases.
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/627171b9-a215-4e91-8986-754346f6e5e2)
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/cb3bf56a-eeb2-47fc-aa77-12a29af947e2)

PR #330, opened 2023-08-15
add county filter on farms endpoint

Omarabdul3ziz:
### Description
add county filter on farms endpoint

### Changes
- add db filter on farms table
- add tests for the new filter
- fix countries slice appending on nodes aggregation test
- update swagger docs

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/329

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/grid-proxy/tests/queries/farm_test.go
+++ b/grid-proxy/tests/queries/farm_test.go
@@ -41,6 +41,11 @@ var farmFilterRandomValueGenerator = map[string]func(agg FarmsAggregate) interfa
 		name := changeCase(agg.farmNames[rand.Intn(len(agg.farmNames))])
 		return &name
 	},
+	"Country": func(_ FarmsAggregate) interface{} {
+		aggNode := calcNodesAggregates(&data)
+		country := changeCase(aggNode.countries[rand.Intn(len(aggNode.countries))])
+		return &country
+	},
 	"NameContains": func(agg FarmsAggregate) interface{} {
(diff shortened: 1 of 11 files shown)


grid-proxy/tests/queries/farm_test.go
```diff
@@ -41,6 +41,11 @@ var farmFilterRandomValueGenerator = map[string]func(agg FarmsAggregate) interfa
 		name := changeCase(agg.farmNames[rand.Intn(len(agg.farmNames))])
 		return &name
 	},
+	"Country": func(_ FarmsAggregate) interface{} {
```
AbdelrahmanElawady, line 44: why not like other fields with FarmsAggregate

AbdelrahmanElawady (approved):
Looks good just a question

grid-proxy/tests/queries/farm_test.go
Omarabdul3ziz: the `agg` variable is not intended to be used inside this function. however, it is still needed to satisfy the function signature. I set it as blank to bypass the warnings

AbdelrahmanElawady: I am just wondering why is it not intended to be used inside the function.

Omarabdul3ziz: the farm table itself doesn't have a country field so we got the possible counties from the node aggregation

Issue #358, opened 2023-08-30
Enhancing generator.go performance


diaabadr:
## Describe the bug

Increasing performance for generating data from `generator.go` as it takes more time.
## To Reproduce

    1. Go to '"grid-proxy'
    2. Run `make db-start` then `make db-fill`

## Expected behavior

Bad performance for generating these data.

PR #359, opened 2023-08-30
refactoring generate.go queries

diaabadr:
### Description

Enhanced query performance in the `generate.go` by refactoring the insertion process. Replaced individual insertions within a loop with a single insert statement.

### Changes

- Refactored all insertion and update queries to be in single SQL statement
- Added function to utils to extract the values of each statement.

### Related Issues

#358

### Checklist

- [X] Tests included
- [X] Build pass
- [X] Documentation
- [X] Code format and docstring

### Execution time comparison based on the provided constant values in the file:
- **Before:** 84.931081115 seconds
- **After:** 0.993061923 seconds

### DIFF
--- a/grid-proxy/tools/db/generate.go
+++ b/grid-proxy/tools/db/generate.go
@@ -29,14 +29,14 @@ const (
 	contractCreatedRatio = .1 // from devnet
 	usedPublicIPsRatio   = .9
 	nodeUpRatio          = .5
-	nodeCount            = 100000
-	farmCount            = 10000
-	normalUsers          = 50000
+	nodeCount            = 1000
+	farmCount            = 100
+	normalUsers          = 2000
 	publicIPCount        = 1000
 	twinCount            = nodeCount + farmCount + normalUsers
-	nodeContractCount    = 100000
-	rentContractCount    = 1000
-	nameContractCount    = 1000
+	nodeContractCount    = 3000
+	rentContractCount    = 100
+	nameContractCount    = 300
 
 	maxContractHRU = 1024 * 1024 * 1024 * 300
 	maxContractSRU = 1024 * 1024 * 1024 * 300
--- a/grid-proxy/tools/db/generate.go
+++ b/grid-proxy/tools/db/generate.go
@@ -82,7 +82,6 @@ func generateTwins(db *sql.DB) error {
 			twin_id:      i,
 			grid_version: 3,
 		}
-
 		twins = append(twins, twin)
 	}
 
@@ -96,7 +95,7 @@ func generateTwins(db *sql.DB) error {
 
 func generatePublicIPs(db *sql.DB) error {
 	var publicIPs []interface{}
-	var nodeContractUpdateValues []string
+	var nodeContracts []interface{}
 
 	for i := uint64(1); i <= publicIPCount; i++ {
 		contract_id := uint64(0)
@@ -109,9 +108,11 @@ func generatePublicIPs(db *sql.DB) error {
 		}
 		ip := randomIPv4()
 		farmID, err := rnd(1, farmCount)
+		
 		if err != nil {
 			return err
 		}
+		
 		public_ip := public_ip{
 			id:          fmt.Sprintf("public-ip-%d", i),
 			gateway:     ip.String(),
@@ -121,7 +122,7 @@ func generatePublicIPs(db *sql.DB) error {
 		}
 
 		publicIPs = append(publicIPs, public_ip)
-		nodeContractUpdateValues = append(nodeContractUpdateValues, fmt.Sprintf("%d", contract_id))
+		nodeContracts = append(nodeContracts, contract_id)
 
 	}
 
@@ -129,13 +130,10 @@ func generatePublicIPs(db *sql.DB) error {
 		return err
 	}
 
-	if len(nodeContractUpdateValues) != 0 {
-		query := "UPDATE node_contract set number_of_public_i_ps = number_of_public_i_ps + 1 WHERE contract_id IN ("
-		query += strings.Join(nodeContractUpdateValues, ",") + ");"
-		if _, err := db.Exec(query); err != nil {
-			return err
-		}
+	if err := updateNodeContractPublicIPs(db, nodeContracts); err != nil {
+		return err
 	}
+
 	fmt.Println("public IPs generated")
 
 	return nil
@@ -143,6 +141,7 @@ func generatePublicIPs(db *sql.DB) error {
 
 func generateFarms(db *sql.DB) error {
 	var farms []interface{}
+
 	for i := uint64(1); i <= farmCount; i++ {
 		farm := farm{
 			id:                fmt.Sprintf("farm-%d", i),
@@ -155,6 +154,7 @@ func generateFarms(db *sql.DB) error {
 			grid_version:      3,
 			stellar_address:   "",
 		}
+
 		if farm.dedicated_farm {
 			dedicatedFarms[farm.farm_id] = struct{}{}
 		}
@@ -181,6 +181,7 @@ func generateNodeContracts(db *sql.DB, billCount *int) error {
 			return err
 		}
 		state := "Deleted"
+
 		if nodeUP[nodeID] {
 			if flip(contractCreatedRatio) {
 				state = "Created"
@@ -188,21 +189,26 @@ func generateNodeContracts(db *sql.DB, billCount *int) error {
 				state = "GracePeriod"
 			}
 		}
+
 		if state != "Deleted" && (minContractHRU > nodesHRU[nodeID] || minContractMRU > nodesMRU[nodeID] || minContractSRU > nodesSRU[nodeID]) {
 			i--
 			continue
 		}
+
 		twinID, err := rnd(1100, 3100)
 		if err != nil {
 			return err
 		}
+
 		if renter, ok := renter[nodeID]; ok {
 			twinID = renter
 		}
+
 		if _, ok := availableRentNodes[nodeID]; ok {
 			i--
 			continue
 		}
+
 		contract := node_contract{
 			id:                    fmt.Sprintf("node-contract-%d", i+rentContractCount),
 			twin_id:               twinID,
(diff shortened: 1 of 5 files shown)


grid-proxy/tools/db/utils.go
```diff
@@ -44,6 +45,15 @@ func max(a, b uint64) uint64 {
 	return b
 }
 
+func extractValuesFromInsertQuery(q string) (string, error) {
```
xmonader: Add test for this function and include docs for the behavior

grid-proxy/tools/db/generate.go
```diff
@@ -70,14 +83,32 @@ func generateTwins(db *sql.DB) error {
 			twin_id:      i,
 			grid_version: 3,
 		}
-		if _, err := db.Exec(insertQuery(&twin)); err != nil {
+
+		value, err := extractValuesFromInsertQuery(insertQuery(&twin))
+		if err != nil {
+			panic(err)
```
xmonader: the function returns the error, and the caller in generateData is acting on that value, so don't panic here, please refactor to return and the call can do whatever.

grid-proxy/tools/db/generate.go
```diff
@@ -361,34 +534,82 @@ func generateNodes(db *sql.DB) error {
...
+			publicConfigValues = append(publicConfigValues)
+
+		}
+	}
+
+	if len(locationValues) != 0 {
+		query := "INSERT INTO location (id, longitude, latitude) VALUES"
+		query += strings.Join(locationValues, ",") + ";"
+		if _, err := db.Exec(query); err != nil {
+			panic(err)
```
xmonader: same comments as in generateTwins, please apply everywhere.

grid-proxy/tools/db/generate.go
```diff
@@ -70,14 +83,32 @@ func generateTwins(db *sql.DB) error {
 			twin_id:      i,
 			grid_version: 3,
 		}
-		if _, err := db.Exec(insertQuery(&twin)); err != nil {
+
+		value, err := extractValuesFromInsertQuery(insertQuery(&twin))
```
mariobassem: i see that insertQuery is only called in extractValuesFromInsertQuery, hence i think it should be called internally if it's needed at all.

grid-proxy/tools/db/generate.go
```diff
@@ -70,14 +83,32 @@ func generateTwins(db *sql.DB) error {
...
+		if err != nil {
+			panic(err)
+		}
+		values = append(values, value)
+	}
+
+	if len(values) != 0 {
```
mariobassem: for better readability, wrap this block in a function whose only purpose is to insert twin records

grid-proxy/tools/db/generate.go
```diff
@@ -91,17 +122,39 @@ func generatePublicIPs(db *sql.DB) error {
...
 			panic(err)
 		}
 	}
+	fmt.Println("public IPs generated")
+
 	return nil
 }
 
 func generateFarms(db *sql.DB) error {
+	var values []string
```
mariobassem: rename to something more descriptive

grid-proxy/tools/db/generate.go
```diff
@@ -117,14 +170,32 @@ func generateFarms(db *sql.DB) error {
...
+		if _, err := db.Exec(query); err != nil {
+			panic(err)
+		}
+	}
+	fmt.Println("Farms generated")
+
 	return nil
 }
 
 func generateContracts(db *sql.DB) error {
```
mariobassem: generate contracts should generate all contracts (node, name, rent). maybe embed the three contract generation functions in it.

mariobassem (changes requested):
great work Diaa. just some small comments

grid-proxy/tools/db/types.go
```diff
@@ -54,6 +54,8 @@ type twin struct {
 	relay        string
 	public_key   string
 }
+
```
mariobassem, line 57: remove new line

grid-proxy/tools/db/utils.go
```diff
@@ -2,21 +2,32 @@ package main
...
 	"fmt"
 	"math/rand"
 	"net"
 	"reflect"
 	"strings"
 )
 
-func rnd(min, max uint64) uint64 {
-	return rand.Uint64()%(max-min+1) + min
+var errMaxMinInvalid = errors.New("min cannot be greater than max")
```
mariobassem: we don't really need to create an error variable, right?
if it's to easily check it when testing, just checking if there is an error is usually enough.

grid-proxy/tools/db/utils_test.go
```diff
@@ -0,0 +1,39 @@
...
+	assert.Equal(t, errMaxMinInvalid, got)
+}
+
+func TestMax(t *testing.T) {
+	want := uint64(5)
+	got := max(4, 5)
+	assert.Equal(t, want, got)
+}
+
+func TestObjectToTupleString(t *testing.T) {
```
mariobassem, line 28: great!
maybe just add one more test run in this function to test NULL values?

grid-proxy/tools/db/generate.go
```diff
@@ -31,10 +31,10 @@ const (
 	nodeUpRatio          = .5
 	nodeCount            = 1000
 	farmCount            = 100
-	normalUsers          = 2000
+	normalUsers          = 200
```
mariobassem: why did normal users count decrease to 200??

grid-proxy/tools/db/generate.go
```diff
@@ -31,10 +31,10 @@ const (
 	nodeUpRatio          = .5
 	nodeCount            = 1000
```
mariobassem: maybe increase the numbers a bit??
let's do nodes 100000, farms 10000, nodeContracts 100000, rentContracts 1000, nameContracts 1000, normalUsers 50000. 
what do you think?

grid-proxy/tools/db/generate.go
```diff
@@ -397,37 +606,113 @@ func generateNodeGPUs(db *sql.DB) error {
...
+	fmt.Println("node GPUs generated")
+
+	return nil
+
+}
+
+func generateContracts(db *sql.DB) error {
+	billCount := 1
+	// Note: it's order senstive
+	if err := generateRentContracts(db, &billCount); err != nil {
```
mariobassem: having to pass the billCount by reference works in our use case, but in general it's not a really good idea unless it's needed, as a pointer creates the possibility for the value to not even be there.
i believe we can work around that, right?

grid-proxy/tools/db/generate.go
```diff
@@ -397,37 +606,113 @@ func generateNodeGPUs(db *sql.DB) error {
...
+
+	fmt.Println("node GPUs generated")
+
+	return nil
+
+}
+
+func generateContracts(db *sql.DB) error {
+	billCount := 1
+	// Note: it's order senstive
```
mariobassem: is there a way to make it not order sensitive?

grid-proxy/tools/db/generate.go
```diff
@@ -70,38 +82,66 @@ func generateTwins(db *sql.DB) error {
...
+		return err
 	}
+	fmt.Println("twins generated")
+
 	return nil
 }
 
 func generatePublicIPs(db *sql.DB) error {
+	var publicIPs []string
+	var nodeContracts []interface{}
```
mariobassem: does nodeContracts need to be []interface{} ?

grid-proxy/tools/db/generate.go
```diff
@@ -70,38 +82,66 @@ func generateTwins(db *sql.DB) error {
...
-			contract_id = createdNodeContracts[rnd(0, uint64(len(createdNodeContracts))-1)]
+			idx, err := rnd(0, uint64(len(createdNodeContracts))-1)
+			if err != nil {
+				return err
+			}
+			contract_id = createdNodeContracts[idx]
 		}
 		ip := randomIPv4()
+		farmID, err := rnd(1, farmCount)
+
```
mariobassem: remove empty line?

grid-proxy/tools/db/generate.go
```diff
@@ -216,47 +324,74 @@ func generateNameContracts(db *sql.DB) error {
...
+	if err := insertTuples(db, contract_bill_report{}, billReports); err != nil {
+		return err, billsStartID, contractsStartID
+	}
+
+	fmt.Println("name contracts generated")
+
+	return nil, billsStartID, contractsStartID
 }
-func generateRentContracts(db *sql.DB) error {
+func generateRentContracts(db *sql.DB, billsStartID, contractsStartID int) (error, int, int) {
```
mariobassem: nothing mandates that an error should have a specific position an a function's returned values, but if we care about our code's readability we would stick with the convention of having the error as the last returned value, right?

grid-proxy/tools/db/generate.go
```diff
@@ -397,37 +609,117 @@ func generateNodeGPUs(db *sql.DB) error {
...
+
+	return nil
+
+}
+
+func generateContracts(db *sql.DB) error {
+	billsStartID := 1
+	contractsStartID := 1
+
+	err, billsStartID, contractCount := generateRentContracts(db, billsStartID, contractsStartID)
```
mariobassem: this is better from last version. 
i would go further with this and make the generateRentContract, generateNodeContracts, and generateNameContracts functions just create the string arrays of the contracts and the bill reports, and do the actual insert calls here in the generateContracts function.
but this is just nitpicking, feel free to ignore.

mariobassem (changes requested):

grid-proxy/tools/db/generate.go
diaabadr: formatter does not remove them.

grid-proxy/tools/db/generate.go
diaabadr: what about generating all contracts in the functions and collecting reports from each one and inserting them once? that's the one which will make the performance better as inserting all bills at once.

grid-proxy/tools/db/generate.go
```diff
@@ -267,72 +399,131 @@ func generateRentContracts(db *sql.DB) error {
...
+
 		up := flip(nodeUpRatio)
-		updatedAt := time.Now().Unix() - int64(rnd(60*40*3, 60*60*24*30*12))
+		periodFromLatestUpdate, err := rnd(60*40*3, 60*60*24*30*12)
+		if err != nil {
+			return err
+		}
+		updatedAt := time.Now().Unix() - int64(periodFromLatestUpdate)
+
+		periodFromLatestUpdate, err = rnd(0, 60*40*1)
```
mariobassem: should this be in the `if up {}` block ??

mariobassem (approved):

mariobassem (approved):

xmonader (approved):

Issue #365, opened 2023-09-01
/nodes performance issue with filters


diaabadr:
## Describe the bug

Bad performance

## URLs that has performance issues:
1. [link](https://gridproxy.grid.tf/nodes?free_mru=8589934592&free_sru=55834574848&country=United%20States&available_for=1061).
2. [link](https://gridproxy.grid.tf/nodes?free_mru=8589934592&free_sru=55834574848&country=United%20States&available_for=1061&status=up)
3. [link](https://gridproxy.grid.tf/nodes?free_mru=8589934592&country=United%20States&available_for=1061)
4. [link](https://gridproxy.grid.tf/nodes?free_mru=8589934592&country=United%20States&available_for=1061&status=up)
5. [link](https://gridproxy.grid.tf/nodes?free_sru=55834574848&country=United%20States&available_for=1061)
6. [link](https://gridproxy.grid.tf/nodes?free_sru=55834574848&country=United%20States&available_for=1061&status=up)
7. [link](https://gridproxy.grid.tf/nodes?country=United%20States&available_for=1061)
8. [link](https://gridproxy.grid.tf/nodes?country=United%20States&available_for=1061&status=up)
## Expected behavior

all of them have performance issues

PR #366, opened 2023-09-01
Improved Nodes Query Performance

diaabadr:
### Description
Changes introduced: Enhancing Performance Through Index Addition

- Initial execution time ranged from 12 to 15 seconds.
- Now the execution time has been reduced to about 120 milliseconds.
**This numbers based on 10,000 node, 1,000 farm, 2,000 normal users, 13,000 twin, 60,000 node_contract, and 1,000 rent_contract.**

### Changes
Adding 2 indices:
- one on rent_contract `node_id`
    > Optimizes join performance for rent_contract table using merge join which is  more effecient than nested loop join. 
- second on node_contract `state`
    > This helps for index scan in state attribute instead of seq scan. but its benefits may not be fully realized for `node_contracts  < 30,000` records. but as the number of records grows, it will be necessary.

### Related Issues
#365


diaabadr:
This is not the best solution for filters. it solved the issue for the url specified in the related issue. but does not resolve the issue for other filters within the same endpoint as [Ex1](https://gridproxy.grid.tf/nodes?free_mru=8589934592&free_sru=55834574848&country=United%20States&available_for=1061), [Ex2] (https://gridproxy.grid.tf/nodes?filter%20free_mru=8589934592&free_sru=55834574848&country=United%20States&total_sru%3E290000)
So i will mark it as a draft  until I can find a more effective solution.

diaabadr:
This is not the best solution for filters. it solved the issue for the url specified in the related issue. but does not resolve the issue for other filters within the same endpoint.
So i will mark it as a draft  until I can find a more effective solution.

diaabadr:
### New changes
- Adding index on farm_id at farm table and index on farm_id on node_table. this provides merge join
- Adding index on node(country). helps with country filters.
- index on node_contract(node_id) to help in join filters.
- restructing some sections in the query.
**Now all URLs mentioned in the Issue should work fine.**

Issue #302, opened 2023-08-08
SQL 53300 state 


sameh-farouk:
## Describe the bug
I noticed this in the Devnet Gridproxy server logs

Not sure, but SQL 53300 state may have something to do with the db connection pooler. if opening these many connections is not a bug in gridproxy, we could then try to increase `max_connections` if the value is too low.

## Screenshots
![Screenshot from 2023-08-08 16-30-32](https://github.com/threefoldtech/tfgrid-sdk-go/assets/42457449/251476c5-dc25-4ae3-afb9-7f56af67263f)

## Update:
The db connection pool maximum open connections should be set (using `SetMaxOpenConns()` method) to a value lower than the `max_connections` parameter in PostgreSQL, which is 100 by default.

See this [comment](https://github.com/threefoldtech/tfgrid-sdk-go/issues/302#issuecomment-1670353846) for more detalis.

A-Harby:
It appears that there is a limitation on the amount of requests that can be made to the grid proxy before it crashes or refuse the request, as shown in the [issue](https://github.com/threefoldtech/tfgrid-sdk-ts/issues/942); this should be fixed from the dashboard side, but it reveals that the grid proxy's stress load isn't the best.

sameh-farouk:
> It appears that there is a limitation on the amount of requests that can be made to the grid proxy before it crashes or refuse the request, as shown in the https://github.com/threefoldtech/tfgrid-sdk-ts/issues/942; this should be fixed from the dashboard side, but it reveals that the grid proxy's stress load isn't the best.

Indeed there is a limitation but in this case it is more about the amount of requests that can be made to the db.
The db configuration sets a concurrent connections limit for PostgreSQL that prevents more connections than the maximum size. The client also set a limit for the number of connections (pool), When the limit is hit, new requests have to wait until the ongoing queries end and give up their connections. the wait will be also limited to some seconds (actual value is depends on how this is configured also) then these pending requests would timeout and gives such a nice error.

Checking our db instance,

```bash
root@processor-db-75b55b9864-2vjmp:/# psql -U postgres -c 'SHOW max_connections;'
 max_connections 
-----------------
 100
(1 row)
```

[...]

PR #371, opened 2023-09-06
add docs for gridproxy database

mariobassem:
### Description

This PR adds a document for the gridproxy database.

### Changes

- New doc file for the database

### Related issues

- #302

### DIFF
--- a/grid-proxy/docs/database.md
+++ b/grid-proxy/docs/database.md
@@ -1,0 +1,11 @@
+# Database
+
+The grid proxy has access to a postgres database containing information about the tfgrid.\
+The grid proxy mainly retrieves information from the db (maybe with a few modifications for efficient retrieval)\
+
+## Max Open Connections
+
+The db can handle 100 open connections concurrently (that is the default value for postgres).\
+The explorer creates a connection pool to the postgres db, with a max open pool connections set to a specific number (currently 80).\
+It's important to distinguish between the database max connections, and the max pool open connections, because if the pool did not have any constraints, it would try to open as many connections as it wanted, without any notion of the maximum connections the database accepts. It's the database responsibility then to accept or deny the connection.\
+This is why the max number of open pool connections is set to 80: It's below the max connections the database could handle (100), and it gives room for other actors outside of the explorer to open connections with the database.\
--- a/grid-proxy/docs/database.md
+++ b/grid-proxy/docs/database.md
@@ -1,11 +1,12 @@
 # Database
 
-The grid proxy has access to a postgres database containing information about the tfgrid.\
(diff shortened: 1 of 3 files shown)


grid-proxy/docs/database.md
```diff
@@ -0,0 +1,11 @@
+# Database
+
+The grid proxy has access to a postgres database containing information about the tfgrid.\
```
xmonader: why the `\` in every line?

xmonader: specific information for the database or the current required storage?

grid-proxy/docs/database.md
```diff
@@ -0,0 +1,11 @@
...
+The grid proxy mainly retrieves information from the db (maybe with a few modifications for efficient retrieval)\
```
xmonader: How is that database is populated, what are the few modifications re we talking about? denormalization, indices, .. etc?

grid-proxy/docs/database.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Max Open Connections
+
+The db can handle 100 open connections concurrently (that is the default value for postgres).\
```
xmonader: that's the max value ? for the proxy or the for the postgres? how can can i change that value?

grid-proxy/docs/database.md
```diff
@@ -0,0 +1,11 @@
...
+The explorer creates a connection pool to the postgres db, with a max open pool connections set to a specific number (currently 80).\
```
xmonader, line 10: Why 80 was chosen? can i configure it?

mariobassem: it is explained in the document why 80 is used [here](https://github.com/threefoldtech/tfgrid-sdk-go/pull/371/files#diff-daa7b575edcdbd3ca9d60f42426b2519378c4db7c4bb5bc31397e066f912cf2dR11)

it is because the max connections limit from postgres is 100, we could set it to 100, but then if there is one open connection from another actor, other than the explorer, the too many connections error may result in the explorer or the other actor, this is why it is set to 80.

mariobassem: i'll add the configuration part

grid-proxy/docs/database.md
mariobassem: this is the default value set by postgres. you can increase it as you want as long as the infrastructure can handle it. 
you can configure the postgres.conf file with the desired values, or execute an update query on this value, but this requires a db restart.

grid-proxy/docs/database.md
mariobassem: > why the `\` in every line?

just to separate lines when rendering the md file.

xmonader (approved):

Issue #387, opened 2023-09-19
(Mainnet verifications) duplicated nodes records


Omarabdul3ziz:
## Describe the bug
caused by the joins to the `node_gpu` table. when there are two cards with the same `node_twin_id`.

## Original issue
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1106

khaledyoussef24:
env used : mainnet
version : 2.2.0

verified issue is no longer reproducible 
tried many cases 
![Screenshot from 2024-01-16 13-13-09](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/035cf3b3-c811-45ac-8c8e-53883bf32a7e)
![Screenshot from 2024-01-16 13-12-56](https://github.com/threefoldtech/tfgrid-sdk-ts/assets/110984055/47ca295e-f7d6-4f47-8384-dc55d6ce739e)

PR #388, opened 2023-09-19
fix duplicated nodes record 

Omarabdul3ziz:
### Description
fix the nodes duplication records by applying the filters on the node_gpu table on a subquery and then inner join this query with the node table.
the join with node_gpu table will only be applied if only one of the GPU filters is applied.

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/387
- should also help in fixing https://github.com/threefoldtech/tf_operations/issues/1908

### DIFF
--- a/grid-proxy/internal/explorer/db/postgres.go
+++ b/grid-proxy/internal/explorer/db/postgres.go
@@ -377,7 +377,7 @@ func (d *PostgresDatabase) nodeTableQuery() *gorm.DB {
 	return d.gormDB.
 		Table("node").
 		Select(
-			"DISTINCT node.id",
+			"node.id",
 			"node.node_id",
 			"node.farm_id",
 			"node.twin_id",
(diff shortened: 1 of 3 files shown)


grid-proxy/internal/explorer/db/postgres.go
```diff
@@ -377,7 +377,7 @@ func (d *PostgresDatabase) nodeTableQuery() *gorm.DB {
 	return d.gormDB.
 		Table("node").
 		Select(
-			"node.id",
+			"DISTINCT node.id",
```
xmonader: please add a comment to explain the need for distinct

mariobassem (approved):

grid-proxy/internal/explorer/db/postgres.go
Omarabdul3ziz: at the end needed to drop that and use a subquery with to node_gpu table and inner join it with the node table.

The problem with distinct was postgres requires a field that is distinctly selected to be the initial value in the order by statement and that works fine but corrupts the `available_for` filter since we have another order statement to have the nodes rented by this twin in the first of the response.

rawdaGastan (approved):

Issue #402, opened 2023-10-08
create a telegrambot  `user-contracts-mon`


xmonader:
should monitor

- contracts in graceperiod
- contracts against down nodes

reporting:
- every 2hrs `configurable`
- should require the user mnemonic to start

Eslam-Nawara:
- add /stop to let a chat unsubscribe from the bot
- add Dockerfile
- refactor monitor and parser code

PR #405, opened 2023-10-09
implement telegram contracts monitoring bot

Eslam-Nawara:
### Description

This pr adds implementation to contracts monitoring bot

### Changes

- create telegram bot `user-contracts-mon`
-  monitor user contracts in `grace period` and `against down node`
- send notifications on fixed intervals

### Related Issues

- #402


user-contract-mon/internal/monitor.go
```diff
@@ -0,0 +1,66 @@
...
+	interval int    `env:"INTERVAL"`
+}
+
+func NewMonitor(envPath string) (Monitor, error) {
+	mon := Monitor{}
+	envMap, err := parseConfig(envPath)
+	if err != nil {
+		return mon, err
+	}
+	for key, value := range envMap {
```
AbdelrahmanElawady: I don't think you need to loop over the map, you can just access it like `envMap[key]`

user-contract-mon/internal/monitor.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	ticker := time.NewTicker(time.Duration(mon.interval) * time.Second)
+	for range ticker.C {
+		contracts, err := tfPluginClient.ContractsGetter.ListContractsByTwinID([]string{"GracePeriod"})
+		if err != nil {
+			return err
+		}
+
+		mon.Bot.SendMessage(fmt.Sprintf("%v", contracts), chatID, nil)
```
AbdelrahmanElawady: It would be better if it is returned to the user in a prettier format instead of a struct with slices. 
Something like that could work but feel free to use something different:
```
contracts in grace period:
- 512
- 1233
- 3459
```

user-contract-mon/main.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"log"
+
+	monitor "github.com/Eslam-Nawara/user-contract-mon/internal"
+	"github.com/NicoNex/echotron/v3"
+)
+
+func main() {
+	mon, err := monitor.NewMonitor("./.config")
```
AbdelrahmanElawady: instead of hardcoding the config file, we can pass it as an argument and use default if not provided. using `flag` package for example

user-contract-mon/main.go
```diff
@@ -0,0 +1,26 @@
...
+	if err != nil {
+		panic("failed to load configrations")
```
AbdelrahmanElawady: you can log the error and exit using `os.exit(1)` for more user friendly approach

user-contract-mon/main.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	for update := range echotron.PollingUpdates(mon.BotToken) {
+		if update.Message.Text == "/start" {
+
+			log.Printf("[%s] %s", update.Message.From.Username, update.Message.Text)
+			err = mon.StartMonitoring(update.ChatID())
+			if err != nil {
+				panic(err)
```
AbdelrahmanElawady: here too

user-contract-mon/internal/parser.go
```diff
@@ -0,0 +1,22 @@
+package monitor
+
+import (
+	"os"
+	"strings"
+
+	env "github.com/hashicorp/go-envparse"
+)
+
+func parseConfig(envPath string) (map[string]string, error) {
```
AbdelrahmanElawady: adding tests for this package would be great

user-contract-mon/internal/monitor.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/NicoNex/echotron/v3"
+	"github.com/rs/zerolog/log"
+	"github.com/threefoldtech/tfgrid-sdk-go/grid-client/deployer"
+)
+
+type Monitor struct {
```
AbdelrahmanElawady: please document exported sturcts and methods

AbdelrahmanElawady (changes requested):
Good Job!
Also:
- add the directory to `go.work` file in root dir
- add Readme
- I think contracts with down nodes still not reported

AbdelrahmanElawady:
also make sure if any of the required configuration is not provided to report what's missing to the user

user-contracts-mon/main
mariobassem, line 1: do we need the binary file?

user-contracts-mon/internal/parser.go
```diff
@@ -0,0 +1,45 @@
...
+	"os"
+	"strconv"
+	"strings"
+
+	env "github.com/hashicorp/go-envparse"
+)
+
+var invalidCfgError = errors.New("Invalid or Missing Fields in configration file")
+
+func parseFile(envPath string) (string, error) {
```
mariobassem: parse file only reads the file's contents, right?
i mean is there any parsing envolved?

user-contracts-mon/internal/parser.go
```diff
@@ -0,0 +1,45 @@
...
+	envContent, err := os.ReadFile(envPath)
+	if err != nil {
+		return "", err
+	}
+
+	return string(envContent), nil
+}
+
+func parseMonitor(envContent string) (Monitor, error) {
```
mariobassem: what if parseMonitor would accept an `io.Reader` ?
i think it would be better in testing ( you could provide whichever type that implements the Reader interface)
it also would also exempt you from creating a reader from the content string (line 26)

user-contracts-mon/internal/parser.go
```diff
@@ -0,0 +1,45 @@
...
+	mon.Network = envMap["NETWORK"]
+
+	interval, err := strconv.Atoi(envMap["INTERVAL"])
+	if err != nil {
+		return Monitor{}, invalidCfgError
+	}
+	mon.interval = interval
+
+	if mon.Mnemonic == "" || mon.Network == "" || mon.BotToken == "" {
+		return Monitor{}, invalidCfgError
```
mariobassem: can you specify which invalid configs are missing? a user won't know

mariobassem (changes requested):
great work Eslam. 
some small comments:

user-contracts-mon/internal/parser.go
```diff
@@ -0,0 +1,55 @@
...
+	"errors"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+
+	env "github.com/hashicorp/go-envparse"
+)
+
+var invalidCfgError = errors.New("Invalid or Missing Fields in configration file")
```
xmonader: lowercase your error messages

user-contracts-mon/internal/parser.go
```diff
@@ -0,0 +1,55 @@
...
+
+func readFile(envPath string) (io.Reader, error) {
```
xmonader: what is the point of this function? also, *os.File implements the reader interface can be used wherever you needed to use that Reader.

user-contracts-mon/internal/parser.go
```diff
@@ -0,0 +1,55 @@
...
+	envFile, err := os.Open(envPath)
+	if err != nil {
+		return strings.NewReader(""), err
+	}
+
+	return envFile, nil
+}
+
+func parseMonitor(envFile io.Reader) (Monitor, error) {
```
xmonader: please refactor this to, parseConfig that returns your config struct and have a `NewMonitor` that takes that config struct as a param

user-contracts-mon/main.go
```diff
@@ -0,0 +1,39 @@
...
+	"log"
+	"os"
+
+	"github.com/NicoNex/echotron/v3"
+	"github.com/threefoldtech/tfgrid-sdk-go/grid-client/deployer"
+	monitor "github.com/threefoldtech/tfgrid-sdk-go/user-contracts-mon/internal"
+)
+
+func main() {
+	env := flag.String("e", "./.env", "Path to env file")
```
xmonader: use StringVar instead of String, don't use a default value here please.

user-contracts-mon/main.go
```diff
@@ -0,0 +1,39 @@
...
+	flag.Parse()
+
+	mon, err := monitor.NewMonitor(*env)
+	if err != nil {
+		log.Printf(err.Error())
+		os.Exit(1)
```
xmonader: can be replaced with log.Fatal

user-contracts-mon/main.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	log.Printf("monitoring bot has started and waiting for requests")
+
+	tfPluginClient, err := deployer.NewTFPluginClient(mon.Mnemonic, "sr25519", mon.Network, "", "", "", 0, true)
+	if err != nil {
+		log.Printf("Failed to establish gird connection")
+		os.Exit(1)
+	}
```
xmonader: same as above

user-contracts-mon/main.go
```diff
@@ -0,0 +1,39 @@
...
+	"flag"
+	"log"
+	"os"
+
+	"github.com/NicoNex/echotron/v3"
+	"github.com/threefoldtech/tfgrid-sdk-go/grid-client/deployer"
+	monitor "github.com/threefoldtech/tfgrid-sdk-go/user-contracts-mon/internal"
+)
+
+func main() {
```
xmonader, line 9: create StartApp function that does creation of the monitor, the tfplugin, and the pulling (and returns the error) and the only code we should have in main is 

if err := StartApp(); err != nil {

    log.Fatal....
}

user-contracts-mon/internal/parser.go
Eslam-Nawara: This function makes it easy to test the parser without the need to create a new env file for every test case

user-contracts-mon/app/app.go
```diff
@@ -0,0 +1,47 @@
...
+	tfPluginClient, err := deployer.NewTFPluginClient(mon.Mnemonic, "sr25519", mon.Network, "", "", "", 0, true)
+	if err != nil {
+		return errors.New("failed to establish gird connection")
+	}
+	log.Printf("grid connection established successfully")
+
+	for update := range echotron.PollingUpdates(mon.BotToken) {
+		if update.Message.Text == "/start" {
+			log.Printf("[%s] %s", update.Message.From.Username, update.Message.Text)
+			go mon.StartMonitoring(tfPluginClient, update.ChatID())
```
AbdelrahmanElawady: I am curious if a user sent `/start` multiple times would it start multiple goroutines? and if so that shouldn't happen, right?

user-contracts-mon/internal/monitor.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	nodeContracts := contracts.NodeContracts
+	nodeContracts = append(nodeContracts, contracts.RentContracts...)
+
+	contractsIds := ""
+	upNodes := make(chan string)
+
+	for _, contract := range nodeContracts {
+		go isNodeUp(tfPluginClient, contract, upNodes)
```
AbdelrahmanElawady: do we need to add `contract := contract` to not let it use the same contract?

user-contracts-mon/internal/parser.go
```diff
@@ -0,0 +1,54 @@
...
+	network  string `env:"NETWORK"`
+	interval int    `env:"INTERVAL"`
+}
+
+func ParseConfig(envPath string) (Config, error) {
+	conf := Config{}
+
+	envFile, err := os.Open(envPath)
+	if err != nil {
+		return conf, errors.New("failed to open config")
```
AbdelrahmanElawady: add the path to the error for easier debugging

user-contracts-mon/internal/monitor.go
```diff
@@ -0,0 +1,146 @@
...
+// NewMonitor creates a new monitor from parsed config/env file
+func NewMonitor(conf Config) (Monitor, error) {
+	mon := Monitor{}
+	mon.Mnemonic = conf.mnemonic
+	mon.Network = conf.network
+	mon.interval = conf.interval
+	mon.BotToken = conf.botToken
+	mon.Bot = echotron.NewAPI(mon.BotToken)
+
+	return mon, nil
```
AbdelrahmanElawady, line 46: it can't return an error, should it return `(Monitor, error)` or just `Monitor`?

user-contracts-mon/internal/parser_test.go
```diff
@@ -0,0 +1,109 @@
...
+		if err != nil {
+			t.Errorf("failed to write to test file")
+		}
+
+		_, err := ParseConfig(testEnv)
+		if err != nil {
+			t.Errorf("parsing should be successful")
+		}
+	})
+	defer os.Remove(testEnv)
```
AbdelrahmanElawady: this should be called immediately after creating the file. you can also use `t.TempDir` to make sure no artifacts created during tests remain after the tests are done.

user-contracts-mon/internal/parser_test.go
```diff
@@ -0,0 +1,109 @@
...
+            NETWORK=   "network"
+            MNEMONIC=  "mnemonic"
+            INTERVAL=  "3"
+		`
+		err = os.WriteFile(testEnv, []byte(env), 0667)
+		if err != nil {
+			t.Errorf("failed to write to test file")
+		}
+
+		_, err := ParseConfig(testEnv)
```
AbdelrahmanElawady: would be great to check the returned struct equals the expected data.

user-contracts-mon/README.md
```diff
@@ -0,0 +1,63 @@
...
+- Monitors user contracts and nodes status.
+- Customizable alerting and notification system.
+
+## Getting Started
+
+### Prerequisites
+
+Ensure that you have installed:
+
+- Go programming language (version 1.16 or higher)
```
AbdelrahmanElawady: the bot is built with go 1.19, does it work with 1.16?

user-contracts-mon/README.md
```diff
@@ -0,0 +1,63 @@
...
+2. Setup your telegram bot and your env
+
+    - Create a new [telegram bot](README.md#create-a-bot) if you don't have.
+    - Create a new env file `.env`, for example:
+
+        ```env
+        BOT_TOKEN=<your bot token>
+        MNEMONIC=<your mnemonics>
+        NETWORK=<main, dev, test, qa>
+        INTERVAL=<number of minutes between notifications>
```
AbdelrahmanElawady: I think in the code interval is using hours not minutes

AbdelrahmanElawady (commented):
Looks great!
couple of questions:
- should we add a Dockerfile for it so we can use Docker directly?
- should we add a `/stop` command to stop it instead of killing the bot process?

user-contracts-mon/internal/monitor.go
Eslam-Nawara: can you explain more what you mean?

AbdelrahmanElawady: there is a problem with using loop variables with goroutines: https://go.dev/blog/loopvar-preview
but your implementation is correct

user-contracts-mon/internal/monitor.go
```diff
@@ -0,0 +1,175 @@
...
+	for {
+		select {
+		case chatID := <-addChatChan:
+			chatIDs[chatID] = true
+
+		case chatID := <-stopChatChan:
+			chatIDs[chatID] = false
+
+		case <-ticker.C:
+			contractsInGracePeriod, contractsAgainstDownNodes, err := runMonitor(tfPluginClient)
```
AbdelrahmanElawady, line 116: error here is not checked

AbdelrahmanElawady: I am not sure the bot will be working with multiple chat IDs, only one right? so no need for a map. also I am not sure if the stop command should just stop the bot from sending messages or stop it from doing anything, like finding contracts. what do you think?

user-contracts-mon/README.md
```diff
@@ -0,0 +1,64 @@
...
+    - Create a new [telegram bot](README.md#create-a-bot) if you don't have.
+    - Create a new env file `.env`, for example:
+
+        ```env
+        BOT_TOKEN=<your bot token>
+        MNEMONIC=<your mnemonics>
+        NETWORK=<main, dev, test, qa>
+        INTERVAL=<number of hours between notifications>
+        ```
+
```
AbdelrahmanElawady, line 39: a section here for running with docker would be great

user-contracts-mon/internal/monitor.go
Eslam-Nawara: It's checked by the sendResponse in the following line

AbdelrahmanElawady (approved):

mariobassem (approved):

Issue #416, opened 2023-10-15
Add Nightly Builds and Tests


AbdelrahmanElawady:
## Description

Trigger test and build workflows nightly for each component.

PR #432, opened 2023-10-25
add nightly build and tests

Eslam-Nawara:
### Description

Schedule all tests and builds to run nightly 

### Related Issues

- #416

### DIFF
--- a/.github/workflows/grid-proxy-build.yml
+++ b/.github/workflows/grid-proxy-build.yml
@@ -4,6 +4,8 @@ on:
   push:
   pull_request:
     types: [opened, reopened, review_requested]
+  schedule:
+    - cron: "25 5 * * *"
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/grid-proxy-build.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
   pull_request:
     types: [opened, reopened, review_requested]
+  schedule:
+    - cron: "25 5 * * *"
```
AbdelrahmanElawady, line 8: I think this workflow publish the docker image too we don't want it published nightly, right?

AbdelrahmanElawady: nvm, this workflow doesn't push the image

AbdelrahmanElawady (approved):

.github/workflows/test.yml
```diff
@@ -2,6 +2,8 @@ name: Test sdk
 
 on:
   push:
+  schedule:
```
AbdelrahmanElawady: as @rawdaGastan mentioned we don't need unit tests in our nightly workflows, the rest is fine

PR #444, opened 2023-10-29
add perf calls

Omarabdul3ziz:
add new perf calls to get the performance tests results

### DIFF
--- a/grid-client/node/node.go
+++ b/grid-client/node/node.go
@@ -374,6 +374,34 @@ func (n *NodeClient) SystemVersion(ctx context.Context) (ver Version, err error)
 	return
 }
 
+// GetPerfTestsResults get all perf tests results
+func (n *NodeClient) GetPerfTestsResults(ctx context.Context) (result string, err error) {
+	ctx, cancel := context.WithTimeout(ctx, n.timeout)
+	defer cancel()
+
+	const cmd = "zos.perf.get_all"
+
+	if err = n.bus.Call(ctx, n.nodeTwin, cmd, nil, &result); err != nil {
+		return
+	}
+
+	return
+}
+
+// GetPerfTestResult get a single perf test result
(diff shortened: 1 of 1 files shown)


grid-client/node/node.go
```diff
@@ -374,6 +374,34 @@ func (n *NodeClient) SystemVersion(ctx context.Context) (ver Version, err error)
 	return
 }
 
+// GetPerfTestsResults get all perf tests results
+func (n *NodeClient) GetPerfTestsResults(ctx context.Context) (result string, err error) {
```
xmonader, line 378: While I don't like naked returns, but I understand you prefered to be consistent with the rest of the module

grid-client/node/node.go
```diff
@@ -374,6 +374,34 @@ func (n *NodeClient) SystemVersion(ctx context.Context) (ver Version, err error)
...
+	ctx, cancel := context.WithTimeout(ctx, n.timeout)
+	defer cancel()
+
+	const cmd = "zos.perf.get_all"
+
+	if err = n.bus.Call(ctx, n.nodeTwin, cmd, nil, &result); err != nil {
```
xmonader: this can be changed to 
```
err = n.bus.Call...

return
```

grid-client/node/node.go
```diff
@@ -374,6 +374,34 @@ func (n *NodeClient) SystemVersion(ctx context.Context) (ver Version, err error)
...
+}
+
+// GetPerfTestResult get a single perf test result
+func (n *NodeClient) GetPerfTestResult(ctx context.Context, testName string) (result string, err error) {
+	ctx, cancel := context.WithTimeout(ctx, n.timeout)
+	defer cancel()
+
+	const cmd = "zos.perf.get"
+
+	if err = n.bus.Call(ctx, n.nodeTwin, cmd, testName, &result); err != nil {
```
xmonader: same change as above

xmonader (approved):

Issue #421, opened 2023-10-17
Add number of the dedicated nodes to the stats endpoint


AhmedHanafy725:
make sure to include the nodes with 0 contracts, the nodes on a dedicated farm, and the rented ones

Omarabdul3ziz:
Work is done:
added the dedicated nodes count to the stats page

Issue #420, opened 2023-10-17
Update the dedicated field on nodes endpoint to be true if the node is free


AhmedHanafy725:
The current dedicated field can be only `true` if the node is in a dedicated farm but if the node is free (can be dedicated in this case), it will be set to `false`.

Omarabdul3ziz:
Work is done:
in the mentioned pr, we updated the logic of deciding if the node is dedicated or not by including nodes that:
- belongs to a dedicated farm
- have no active node contracts
- have an active rent contract

PR #452, opened 2023-11-01
modify the node dedicated field/filter

Omarabdul3ziz:
### Description
1. modify the dedicated node filter/field. node will be dedicated if one of three conditions is true:
- belongs to dedicated farm
- have no node contract
- have a rent contract (either have or don't have node contract)

2. added dedicated nodes count to the `/stats` endpoint

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/420
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/421

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/grid-proxy/internal/explorer/converters.go
+++ b/grid-proxy/internal/explorer/converters.go
@@ -58,7 +58,7 @@ func nodeFromDBNode(info db.Node) types.Node {
 		ExtraFee:          info.ExtraFee,
 	}
 	node.Status = nodestatus.DecideNodeStatus(node.Power, node.UpdatedAt)
-	node.Dedicated = info.FarmDedicated || info.NodeContractCount == 0 || info.RentContractID != 0
+	node.Dedicated = info.FarmDedicated || !info.HasNodeContract || info.RentContractID != 0
 	return node
 }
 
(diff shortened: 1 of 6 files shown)


xmonader:
I don't understand this part 

"have a rent contract (either have or don't have rent contract)"

Omarabdul3ziz:
if we only consider a node is dedicated if the farm is dedicated or the node has no node contract, this will not include the already rented contract where the renter created a node contract on it. so in the case of a rented node (empty or has deployments) will labeled as a dedicated

Edit:
okay got the typo, i meant either have or don't have node contract :D

xmonader:
thanks ya batta(tta)

grid-proxy/internal/explorer/converters.go
```diff
@@ -59,6 +58,7 @@ func nodeFromDBNode(info db.Node) types.Node {
 		ExtraFee:          info.ExtraFee,
 	}
 	node.Status = nodestatus.DecideNodeStatus(node.Power, node.UpdatedAt)
+	node.Dedicated = info.FarmDedicated || info.NodeContractCount == 0 || info.RentContractID != 0
```
AbdelrahmanElawady: I am curious on why having no node contracts make the node dedicated? are newly registered nodes with no contracts considered dedicated or am I missing dedicated meaning? :)

Omarabdul3ziz: At first you needed to mark your farms as dedicated to make your nodes available to rent. but this changed and if there are any node that is completely free and have no contracts it can marked as dedicated and be available to rent.
so yes newly registered nodes will be dedicated

AbdelrahmanElawady: so dedicated means available to be rented? but that means rented nodes (third condition) are not available to be rented, right?

Omarabdul3ziz: dedicated should include both:
- `rentable`: nodes in dedicated farms, or empty nodes.
- `rented`: nodes already rented. (it have/don't have node-contracts. anyway it will be considered as dedicated)

AbdelrahmanElawady (approved):

Issue #447, opened 2023-10-30
The `randomize` query is already supported by gridproxy but not mentioned in Swagger docs


AhmedHanafy725:

Omarabdul3ziz:
yes, noticed that and added it in this pr, https://github.com/threefoldtech/tfgrid-sdk-go/pull/459

khaledyoussef24:
verified .
it can be found in the swagger now
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/51d22a07-efdc-46a7-a117-810a42c4a7d6)

Issue #350, opened 2023-08-21
Support sorting for nodes and farms 


0oM4R:
As gridproxy supports pagination we have to use external sorting as well, not sure what is the required sorting keys but the current ones in dashboard for the farms is the farm id and farm name only 
- #349 
- https://github.com/threefoldtech/tfgrid-sdk-ts/issues/45

Omarabdul3ziz:
work is done: pr in review
add two new params:
- `sort_by`: that can take any of the node/farm field
- `sort_order`: that can be `desc` or `asc`. `asc` is the default

khaledyoussef24:
verified now you can sort nodes and farm by order (ascending by default).
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/5976a181-96f7-4017-bfe5-7e4e67c5ff7b)
desc
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/2d3a3299-b0dd-47ff-8440-323d93208f6a)

swagger also updated
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/54e758d3-4ea7-491e-9d71-b889e9a21e44)


testlodge:
https://app.testlodge.com/a/26076/projects/40893/suites/211680?expand_section=354530#case_4157422

Issue #349, opened 2023-08-21
support sorting contracts


0oM4R:
In playground we are making 3 requests to get the contracts using graphql, we can do only one request to gridproxy and get all the contracts we need, but as the gridproxy support pagination, we need to support sorting as well. we can't support sorting in Vue while we have external sorting.
the current situation in playground is working fine but in case the user have to many contract the sorting may not preforming well  
- #324
## Describe the solution you'd like

support sorting by: 
- ID
- Type
- State"
- Billing Rate
- Solution Type
- Solution Name
- Created At
- Expiration
- Details

Omarabdul3ziz:
work is done: pr in review
add two new params:
- `sort_by`: that can take any of the contract field
- `sort_order`: that can be `desc` or `asc`. `desc` is the default

muhamadazmy:
@Omarabdul3ziz please check the review on your PR

khaledyoussef24:
verified 
env used : devnet
sorting now works with contracts supporting

1. ID
2. Type
3. State
4. Billing Rate
5. Solution Type
6. Solution Name
7. Created At
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/56f4a537-4736-47b7-8ea1-bbb09601d58f)

[Screencast from 12-21-2023 03:26:26 PM.webm](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/840fca2a-b6f8-4362-9b84-16c6a084677d)

test lodge : https://app.testlodge.com/a/26076/projects/40893/suites/234374?expand_section=422331#case_4160543

Issue #243, opened 2023-07-13
Add sorted_by or ordered_by query param


Mahmoud-Emad:
If I have N nodes and want to retrieve all nodes without a status filter, it would be useful to add an `ordered_by` filter key that accepts any field from the node fields. For example, `sorted_by=status__up` or `ordered_by=sru__15`.

I want this feature to improve the user experience as I wrote in https://github.com/threefoldtech/tfgrid-sdk-ts/issues/770#issuecomment-1633754486

xmonader:
This can be addressed next week

Omarabdul3ziz:
related https://github.com/threefoldtech/tfgrid-sdk-go/issues/350

Omarabdul3ziz:
work is done: pr in review
add two new params for the four endpoints `farms`, `nodes`, `twins`, `contracts`:
- `sort_by`: that can take any of the response field name like `/twins?sort_by=relay`, `/nodes?sort_by=free_mru`, `/farms?sort_by=publicIps`
- `sort_order`: that can be `desc` or `asc`. `asc` is the default

khaledyoussef24:
verified 
enviroment used : devnet
![Screenshot from 2023-12-20 20-01-09](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/ffeebedb-14cc-4b0f-b790-16dc3ac54512)
![Screenshot from 2023-12-20 20-01-01](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/aea4ed1f-2150-4c48-b663-382080a7a9ff)

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/6dd2d3b6-e803-4fee-832b-3bab5e71c2e6)
![Screenshot from 2023-12-21 08-41-29](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/6abc580e-495a-4681-9a77-a1eca195b185)
![Screenshot from 2023-12-21 08-41-00](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/c27aec94-33f3-4b52-bb5b-d57ef82a7a65)

testlodge : 
https://app.testlodge.com/a/26076/projects/40893/suites/211680?expand_section=354530#case_4160010

PR #459, opened 2023-11-05
add sorting by query params for the endpoints

Omarabdul3ziz:
### Description
- introduce two new params `sort_by` and `sort_order`
- update the sorting mechanism for farms/nodes/twins/contracts
- update swagger docs

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/349
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/350
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/243
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/53
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/447
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/473

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


grid-proxy/docs/swagger.json
```diff
@@ -44,6 +44,35 @@
...
+                        "name": "randomize",
+                        "in": "query"
+                    },
+                    {
+                        "enum": [
+                            "twin_id",
+                            "node_id",
+                            "type",
+                            "status",
+                            "..."
```
muhamadazmy: what does the `...` mean ? and if it means any why do we use enum in the first place. I think it's better if we specify a static list of possible sort fields instead of allow of passing any value that might not be defined

grid-proxy/docs/swagger.json
```diff
@@ -458,6 +487,34 @@
...
+                        "description": "Get random patch of gateways",
+                        "name": "randomize",
+                        "in": "query"
+                    },
+                    {
+                        "enum": [
+                            "free_mru",
+                            "total_cru",
+                            "free_ips",
+                            "..."
```
muhamadazmy: same

grid-proxy/pkg/types/types.go
```diff
@@ -79,6 +79,8 @@ type Limit struct {
 	Page      uint64
 	RetCount  bool
 	Randomize bool
+	SortBy    string
+	SortOrder string
```
muhamadazmy: I don't like that there is absolutely no validation on the values of these and then are immediately used in sql statements. (if there are any validation please point me to it)

Beside sql injection, accepting any value as input from the user is generally wrong.

Instead, i would define 2 types 
```go
type SortBy string

const (
   AscSort SortBy = "ASC"
   DescSort SrotBy = "DESC"
)

func(s SortBy) Valid()error {
   // if the s value is not exactly one of the possible values return an error.
}
```

Then the `limit` structure should have it's own Valid() method that validate the input AND call valid() on any internal attribute that needs validation.

Then the Limit structure need to always be validate the moment it's parsed from user input. (query params)

grid-proxy/internal/explorer/helpers.go
```diff
@@ -49,6 +49,11 @@ func getLimit(r *http.Request) (types.Limit, error) {
 	size := r.URL.Query().Get("size")
 	retCount := r.URL.Query().Get("ret_count")
 	randomize := r.URL.Query().Get("randomize")
+	sortBy := r.URL.Query().Get("sort_by")
+	sortOrder := r.URL.Query().Get("sort_order")
+
+	limit.SortBy = sortBy
+	limit.SortOrder = sortOrder
```
muhamadazmy: Please check comment regarding user input validation.

grid-proxy/internal/explorer/db/postgres.go
```diff
@@ -837,14 +871,23 @@ func (d *PostgresDatabase) GetContracts(ctx context.Context, filter types.Contra
...
 	if limit.Randomize {
-		q = q.Limit(int(limit.Size)).
-			Offset(rand.Intn(int(count) - int(limit.Size)))
+		q = q.Order("random()")
+	} else if limit.SortBy != "" {
+		order := sortingOrder.Desc
+		if strings.ToUpper(limit.SortOrder) == sortingOrder.Asc {
+			order = sortingOrder.Asc
+		}
+		q = q.Order(fmt.Sprintf("%s %s", limit.SortBy, order))
```
muhamadazmy, line 914: SQL INJECTION

muhamadazmy (changes requested):

Omarabdul3ziz:
need a rebase and merge with the schema pr here https://github.com/threefoldtech/tfgrid-sdk-go/pull/315

Omarabdul3ziz:
only added the validations for the sorting params values, other params validation will be done in another pr, opened issue here https://github.com/threefoldtech/tfgrid-sdk-go/issues/461

muhamadazmy (approved):

Issue #457, opened 2023-11-05
VMs get assigned the same IP on different deployments


AbdelrahmanElawady:
## Describe the bug

Deploying multiple VMs on the same network but different deployments assign the same private IP to the VMs.

## To Reproduce

- Create two deployments with one VM
- Assign the deployments to a network
- Deploy network then deployments
- Check the private IPs of the VMs it should match

## Expected behavior

On a private network VMs shouldn't have the same IP even in different deployments

## Related issue

- https://github.com/threefoldtech/terraform-provider-grid/issues/781

rawdaGastan:
Tested, the deployments now got assigned with different private IPs on the same network

- case1: same network different nodes:
```console
IP1: 10.20.2.2
IP2: 10.20.3.3
```

- case2: same network same node:
```console
IP1: 10.20.2.2
IP2: 10.20.2.3
```

also solves: 
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/478

problem:
- for case1 the IP should be `10.20.3.2` not `10.20.3.3`

issue:
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/545

PR #460, opened 2023-11-05
Ensure network state is updated after operating on deployments

AbdelrahmanElawady:
### Description

Fix a bug where used IPs are not reflected in separate deployment.

### Changes

- Ensure after any operation on deployments (deploy/update/cancel) that the network state is updated with the used IPs.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/457

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/grid-client/deployer/deployment_deployer.go
+++ b/grid-client/deployer/deployment_deployer.go
@@ -29,7 +29,7 @@ func NewDeploymentDeployer(tfPluginClient *TFPluginClient) DeploymentDeployer {
 }
 
 // GenerateVersionlessDeployments generates a new deployment without a version
-func (d *DeploymentDeployer) GenerateVersionlessDeployments(ctx context.Context, dl *workloads.Deployment, usedHosts []byte) (map[uint32]gridtypes.Deployment, error) {
+func (d *DeploymentDeployer) GenerateVersionlessDeployments(ctx context.Context, dl *workloads.Deployment, usedHosts *[]byte) (map[uint32]gridtypes.Deployment, error) {
 	newDl := workloads.NewGridDeployment(d.tfPluginClient.TwinID, []gridtypes.Workload{})
 	err := d.assignNodesIPs(dl, usedHosts)
 	if err != nil {
@@ -71,7 +71,8 @@ func (d *DeploymentDeployer) Deploy(ctx context.Context, dl *workloads.Deploymen
 	newDeploymentsSolutionProvider := map[uint32]*uint64{dl.NodeID: dl.SolutionProvider}
 
 	network := d.tfPluginClient.State.Networks.GetNetwork(dl.NetworkName)
-	newDeployments, err := d.GenerateVersionlessDeployments(ctx, dl, network.GetUsedNetworkHostIDs(dl.NodeID))
+	usedIPs := network.GetUsedNetworkHostIDs(dl.NodeID)
+	newDeployments, err := d.GenerateVersionlessDeployments(ctx, dl, &usedIPs)
 	if err != nil {
 		return errors.Wrap(err, "could not generate deployments data")
 	}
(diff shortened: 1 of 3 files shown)


grid-client/deployer/deployment_deployer.go
```diff
@@ -92,13 +97,17 @@ func (d *DeploymentDeployer) Deploy(ctx context.Context, dl *workloads.Deploymen
...
 	for _, dl := range dls {
 		if err := d.Validate(ctx, dl); err != nil {
 			return err
 		}
-
-		generatedDls, err := d.GenerateVersionlessDeployments(ctx, dl)
+		if _, ok := networkUsedIPs[dl.NetworkName]; !ok {
+			network := d.tfPluginClient.State.Networks.GetNetwork(dl.NetworkName)
+			usedIPs := network.GetUsedNetworkHostIDs(dl.NodeID)
+			networkUsedIPs[dl.NetworkName] = &usedIPs
```
muhamadazmy: why do u need the networkUserIPs map ? and not use the usedIPs directly inside the loop to pass it to the GenerateVersionlessDeployments function ?

grid-client/deployer/deployment_deployer.go
```diff
@@ -28,9 +29,9 @@ func NewDeploymentDeployer(tfPluginClient *TFPluginClient) DeploymentDeployer {
 }
 
 // GenerateVersionlessDeployments generates a new deployment without a version
-func (d *DeploymentDeployer) GenerateVersionlessDeployments(ctx context.Context, dl *workloads.Deployment) (map[uint32]gridtypes.Deployment, error) {
+func (d *DeploymentDeployer) GenerateVersionlessDeployments(ctx context.Context, dl *workloads.Deployment, usedHosts *[]byte) (map[uint32]gridtypes.Deployment, error) {
```
muhamadazmy: I was wondering why you passing a `pointer to an array` but i realized you do this to `modify` the array. So it works as an output.

But why not return the array instead of using this output pattern ?

grid-client/deployer/deployment_deployer.go
```diff
@@ -266,12 +291,10 @@ func (d *DeploymentDeployer) Validate(ctx context.Context, dl *workloads.Deploym
 	return dl.Validate()
 }
 
-func (d *DeploymentDeployer) assignNodesIPs(dl *workloads.Deployment) error {
+func (d *DeploymentDeployer) assignNodesIPs(dl *workloads.Deployment, usedHosts *[]byte) error {
```
muhamadazmy: Why not return the array instead of using this 

```go
fn modeifyArray(input []byte) []byte {
  input = appen(input, 0)
  return input
}
```
It's both cleaner to read and implement

grid-client/deployer/deployment_deployer.go
muhamadazmy: I am also wondering what a `[]byte` represents here. there are no docs on what the inputs are? not the type is descriptive (no type alias with a descriptive name)?

muhamadazmy (changes requested):

grid-client/deployer/deployment_deployer.go
AbdelrahmanElawady: because BatchDeploy takes multiple deployments that might be in different networks so we need to track which used IPs in which network

grid-client/deployer/deployment_deployer.go
AbdelrahmanElawady: It's the last byte in the IP so for 1.1.1.2 we just store 2, I personally do not find the best way to store that but this is how it's used throughout the grid client. We were considering changing it to something like BitSet but I believe this will need some refactoring outside this pr scope

xmonader (approved):

rawdaGastan (approved):

muhamadazmy (approved):
While this code might be working but I noticed the following flows:
- The code is hard to follow, mainly because of lack of strong types (passing []byte) instead of a type with a good descriptive name
- I see stuff like `workloads.Contains` which is confusing because it sounds like if u check if workloads contains something. but in the matter of fact it can be just replaced with the built in function `slices.Contains` 
- Obviously there is a problem with IP assignment because it assumes that all VMs live in the same subnet, which is not true. 
 - The wireguard network is a `/16` range. each zos node then hosts a `/16` range so a host ID is actually `uint16` (last 2 octets of the IP) not `uint8/byte`.

I really think the code can be rewritten to be more abstract, separate and readable. I would not prevent this from merging thought since it's needed to fix an issue.

AbdelrahmanElawady:
> I really think the code can be rewritten to be more abstract, separate and readable. I would not prevent this from merging thought since it's needed to fix an issue.

I agree with the points you made it needs some refactoring and better utilization for the available IPs on the network. I will open some issues to track the required changes.

Issue #462, opened 2023-11-06
Add ips of kubernetes worker after deploying in CLI


A-Harby:
## Describe the bug

Can't we also show the ygg ip of the workers? It would also help the user to show that it was successful for both master and worker, and here are the ips of them.

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/37d1bb66-5b74-44db-be62-3de2ed96f372)

Eslam-Nawara:
## Work Completed:
- update the cli to show all master and workers available ips

AbdelrahmanElawady:
Verified and the workers IPs are included with the output:
```console
$ tf-grid-cli deploy kubernetes -n examplektest11 --ssh ~/.ssh/id_rsa.pub --workers-number 2      
1:41PM INF starting peer session=tf-437467 twin=192
1:41PM INF deploying network
1:42PM INF deploying cluster
1:42PM INF master wireguard ip: 10.20.2.2
1:42PM INF master yggdrasil ip: 300:d205:aee5:1d1e:680c:7abf:e9ff:c55d
1:42PM INF worker1 wireguard ip: 10.20.3.3
1:42PM INF worker0 wireguard ip: 10.20.3.2
1:42PM INF worker1 yggdrasil ip: 301:b989:8977:ff23:593b:d90c:7417:30de
1:42PM INF worker0 yggdrasil ip: 301:b989:8977:ff23:5615:17fc:1220:e0d5
```

PR #471, opened 2023-11-08
show k8s workers ips after deployment with cli

Eslam-Nawara:
### Description

- show workers ygg ips after deployment with grid-cli

### Changes

- show workers ips if they are set to have a ygg ips

### Related Issues

- #462


xmonader:
I think workers can have public IPs as well right? and IPv6 too?
And a wireguard one xD

Eslam-Nawara:
> I think workers can have public IPs as well right? and IPv6 too? And a wireguard one xD

of course they can, but the issue only specified ygg ip, fa should public ips also be shown?

xmonader:
> > I think workers can have public IPs as well right? and IPv6 too? And a wireguard one xD
> 
> of course they can, but the issue only specified ygg ip, fa should public ips also be shown?

Yup, please do

xmonader:
I think we are still missing the wireguard IPs 🙈

Eslam-Nawara:
> I think we are still missing the wireguard IPs 🙈

I can't find it actually :)

rawdaGastan:
@Eslam-Nawara you can find it as `IP` in the k8s node struct (the IP is the private wireguard IP)
- https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/grid-client/workloads/k8s.go#L31

grid-cli/cmd/deploy_kubernetes.go
```diff
@@ -196,6 +198,25 @@ var deployKubernetesCmd = &cobra.Command{
 		if ygg {
 			log.Info().Msgf("master yggdrasil ip: %s", cluster.Master.YggIP)
 		}
+
+		for i, worker := range cluster.Workers {
+			log.Info().Msgf("worker%d wireguard ip: %s", i, worker.IP)
```
rawdaGastan: we can ignore `i` and use worker name

xmonader:
Still one more ip (ipv6) of the master we should all of them like we did for the workers

Eslam-Nawara:
It exists here 
- https://github.com/threefoldtech/tfgrid-sdk-go/pull/471/files#diff-6f37cc47e9c5a93162368b6c661fd85de13526e61554eea42aee73f887a343ecR195

xmonader (approved):

PR #482, opened 2023-11-14
define the result type for perf calls

Omarabdul3ziz:
### Description
define the result type for perf calls

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


grid-client/node/node.go
```diff
@@ -374,8 +374,15 @@ func (n *NodeClient) SystemVersion(ctx context.Context) (ver Version, err error)
 	return
 }
 
+// TaskResult holds the perf test result
+type TaskResult struct {
+	Name      string      `json:"name"`
+	Timestamp uint64      `json:"timestamp"`
```
AbdelrahmanElawady: there is a description field missing here, maybe we should imported from zos directly so we don't have to maintain two identical structs

AbdelrahmanElawady (changes requested):

grid-client/node/node.go
AbdelrahmanElawady: nvm we will not import the type :)
adding the missing field is enough

AbdelrahmanElawady (approved):

Issue #489, opened 2023-11-16
RMB direct client panic: send on closed channel


Eslam-Nawara:
## Describe the bug

When trying to filter 1000 nodes, the RMB direct client panic with `send on closed channel` error
the error happened because the response channel gets closed but remains existing in the response map, so when the client check if the response exists before it listen to it, it finds that the response exists, but with closed channel so it panics.

## Screenshots
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/67752395/90ba64b0-6eca-4b9f-bdeb-2c2d6d63601e)

## Suggestions 
- delete the response object of the responses map when the channel is closed [here](https://github.com/threefoldtech/tfgrid-sdk-go/blob/b1ac1f1e83635d8ebe4a704892aa16587d71261f/rmb-sdk-go/direct/rpc.go#L73)
- log that the channel is closed so the response will be skipped [here](https://github.com/threefoldtech/tfgrid-sdk-go/blob/b1ac1f1e83635d8ebe4a704892aa16587d71261f/rmb-sdk-go/direct/rpc.go#L58)

rawdaGastan:
Verified.

Tried to make 1000 calls to the relay and they passed
```
for i := 0; i < 1000; i++ {
	var ver version
	if err := client.Call(ctx, 23, "zos.system.version", nil, &ver); err != nil {
		return err
	}

	fmt.Printf("%d- output: %s\n", i, ver)
}
```

and it printed the 1000 outputs
```
...
...
995- output: {0.0.0-83df013 v0.2.11}
996- output: {0.0.0-83df013 v0.2.11}
997- output: {0.0.0-83df013 v0.2.11}
998- output: {0.0.0-83df013 v0.2.11}
999- output: {0.0.0-83df013 v0.2.11}
```

PR #490, opened 2023-11-16
fix RMB direct client panic

Eslam-Nawara:
### Description

delete response object from response map after its channel was closed  

### Related Issues

- #489


rawdaGastan (approved):

Eslam-Nawara:
## Work Completed
I've tested the client with 2000 requests using the rpc client example and it worked fine

Issue #496, opened 2023-11-20
FilterNodes returns misleading error


Eslam-Nawara:
## Describe the bug
- FilterNodes returns an error ` could not find any node with free ssd pools: 15 GB` with a nodeID = 39 with 98% free ssd

## To Reproduce

try to use the grid-client to filter the nodes of mainnet with node filter with this [example](https://gist.github.com/Eslam-Nawara/23ee7f9b604c2d8195f613f21c215d9f)

## Expected behavior

expected to get one node with id 39 as it has enough storage, but it fails due to no enough storage

## Screenshots

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/67752395/acb67830-0505-4fa9-a2f8-3ad0990578dc)

Eslam-Nawara:
In this case, the FilterNodes results with that error due to failure in HDD filtration, although the example doesn't use HDD storage at all, but it doesn't check if there's no storage required and starts checking if the requested pools are available in the node and reports failure as there's no available pools even if the required storage is none.

Eslam-Nawara:
Work Completed: 
fixed the filtration problem and changed the error message to be more specific of what might caused the problem

rawdaGastan:
Verified and fixed

Issue #491, opened 2023-11-16
Should we check HDD disks with the size of SSDs?


Eslam-Nawara:
## Describe the bug

NodeFilter function uses the size of SSD disks as a filter for the hdd disks [here](https://github.com/threefoldtech/tfgrid-sdk-go/blob/b1ac1f1e83635d8ebe4a704892aa16587d71261f/grid-client/deployer/node_filter.go#L61) but I think we should use the HDD disks size

rawdaGastan:
Verified, fixed here:
- https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/grid-client/deployer/node_filter.go#L76

PR #498, opened 2023-11-20
fix grid-client hdd filtration

Eslam-Nawara:
### Description

- Fix the filtration function to check if the required storage is none and passes the filtration if so.
- Change the error message if the filtration failed to include the HDD storage and avoid panics on nil FreeSRU or FreeHRU

### Changes

- Changed the FilterNodes function, fixed the call to hasEnoughStorage with HDD, and changed the error message of filure to find node with the required pools.
- Changed the hasEnoughStorage function to skip the check if there's no required storage.

### Related Issues

- #496 
- #491

### DIFF
--- a/grid-client/deployer/node_filter.go
+++ b/grid-client/deployer/node_filter.go
@@ -58,13 +58,22 @@ func FilterNodes(ctx context.Context, tfPlugin TFPluginClient, options types.Nod
 			workloads.Delete(nodePools, node)
 			continue
 		}
-		if hasEnoughStorage(pools, ssdDisks, zos.SSDDevice) && hasEnoughStorage(pools, ssdDisks, zos.HDDDevice) {
+		if hasEnoughStorage(pools, ssdDisks, zos.SSDDevice) && hasEnoughStorage(pools, hddDisks, zos.HDDDevice) {
 			nodePools = append(nodePools, node)
 		}
 	}
 
 	if len(nodePools) == 0 {
-		return []types.Node{}, errors.Errorf("could not find any node with free ssd pools: %d GB", convertBytesToGB(*options.FreeSRU))
+		freeSRU := uint64(0)
+		if options.FreeSRU != nil {
+			freeSRU = convertBytesToGB(*options.FreeSRU)
+		}
+		freeHRU := uint64(0)
+		if options.FreeHRU != nil {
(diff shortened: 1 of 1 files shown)


grid-client/deployer/node_filter.go
```diff
@@ -58,13 +58,22 @@ func FilterNodes(ctx context.Context, tfPlugin TFPluginClient, options types.Nod
...
 		}
-		if hasEnoughStorage(pools, ssdDisks, zos.SSDDevice) && hasEnoughStorage(pools, ssdDisks, zos.HDDDevice) {
+		if hasEnoughStorage(pools, ssdDisks, zos.SSDDevice) && hasEnoughStorage(pools, hddDisks, zos.HDDDevice) {
 			nodePools = append(nodePools, node)
 		}
 	}
 
 	if len(nodePools) == 0 {
-		return []types.Node{}, errors.Errorf("could not find any node with free ssd pools: %d GB", convertBytesToGB(*options.FreeSRU))
+		freeSRU := uint64(0)
```
rawdaGastan: why not 
`var freeSRU uint64`
`var freeHRU uint64`

grid-client/deployer/node_filter.go
```diff
@@ -58,13 +58,22 @@ func FilterNodes(ctx context.Context, tfPlugin TFPluginClient, options types.Nod
...
+		if options.FreeSRU != nil {
+			freeSRU = convertBytesToGB(*options.FreeSRU)
+		}
+		freeHRU := uint64(0)
+		if options.FreeHRU != nil {
+			freeHRU = convertBytesToGB(*options.FreeHRU)
+		}
+
+		return []types.Node{}, errors.Errorf("could not find any node with free ssd pools: %d GB and free hdd pools: %d GB", freeSRU, freeHRU)
```
rawdaGastan, line 82: what if the problem is with ssds or hdds only?

rawdaGastan (changes requested):

grid-client/deployer/node_filter.go
Eslam-Nawara: The error message indicates that there are no nodes with the required storage options available
at this point, we can't tell for sure if the problem was the ssd or the hdd

rawdaGastan (approved):

Issue #497, opened 2023-11-20
Use concurrency in FilterNodes


Eslam-Nawara:
I think it would be much better for a large scale filtration using FilterNodes function to check for the nodes pools concurrently, It would save unnecessary time as every check on every node takes sometime, added the large number of nodes would take a lot of unnecessary time.

Eslam-Nawara:
Work Completed: 

- I've used a function to tests if every node has enough storage pools and tested it on filtration of 1000 nodes and is working fine and much faster

AbdelrahmanElawady:
Verified and there is significant time improvement between using concurrency and not using it. Example tested on 0.11.9 and 0.11.7
```go
func main() {
	now := time.Now()
	tf, err := deployer.NewTFPluginClient(mnemonic, "sr25519", network, "", "", "", 0, true)
	if err != nil {
		panic(err)
	}
	nodes, err := deployer.FilterNodes(context.Background(), tf, types.NodeFilter{Status: &status, FreeSRU: &freeSRU}, []uint64{freeSRU}, nil, nil)
	if err != nil {
		panic(err)
	}
	fmt.Println(len(nodes))
	fmt.Println(time.Since(now))
}
```
on 0.11.7 time is `2m48.045638422s` and on 0.11.9 `12.256152623s`

PR #501, opened 2023-11-20
check for node pools concurrently

Eslam-Nawara:
### Description

check for node pools concurrently to save time in case of large filtration

### Related Issues

- #497


grid-client/deployer/node_filter.go
```diff
@@ -48,26 +49,46 @@ func FilterNodes(ctx context.Context, tfPlugin TFPluginClient, options types.Nod
...
+		go func(node types.Node) {
+			defer wg.Done()
+
+			client, err := tfPlugin.NcPool.GetNodeClient(tfPlugin.SubstrateConn, uint32(node.NodeID))
+			if err != nil {
+				log.Error().Err(err).Msgf("failed to get node '%d' client", node.NodeID)
+
+				lock.Lock()
+				workloads.Delete(nodePools, node)
+				lock.Unlock()
```
AbdelrahmanElawady: nodePools should only have the working nodes, we don't need the delete parts

Eslam-Nawara: I discussed it with rawda, and we didn't know why it was implemented that way with delete fa I left it as it is for now, bs it appears like there's no need for it ah

rawdaGastan: You can remove it 👍🏻

AbdelrahmanElawady (approved):

grid-client/deployer/node_filter.go
```diff
@@ -48,26 +48,38 @@ func FilterNodes(ctx context.Context, tfPlugin TFPluginClient, options types.Nod
...
-	}
+	for _, node := range nodes {
+		wg.Add(1)
+
+		go func(node types.Node) {
+			defer wg.Done()
+
+			client, err := tfPlugin.NcPool.GetNodeClient(tfPlugin.SubstrateConn, uint32(node.NodeID))
+			if err != nil {
+				log.Error().Err(err).Msgf("failed to get node '%d' client", node.NodeID)
```
rawdaGastan: can we change the log level to debug (we don't want the users to get error msgs of not filtered nodes
apply for the ext condition too

rawdaGastan (approved):

PR #504, opened 2023-11-21
rmb peer router for server side implementation

### LINKED ISSUE #499: [rmb] router for direct client to allow building server
Similar to how HTTP MUX does it, we need a router structure that can be used to handle requests from remote users and then completely able to build an app that works as "server" without relying on `rmb-peer` 

We already have something similar at https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/rmb-sdk-go/rmb.go#L133 BUT this router is built to work with `rmb-peer` as mentioned before. 

The new router should then work with the direct client. The direct client establishes a connection and then forward all received messages to a single entry point (handler). The router then can expose this single handler entry point then just route all requests to the proper handler, or return an unknown command (404) similar to http server implementations

for example
```go
       router := rmb.NewRouter()

	app := router.Subroute("calculator")

	// this function then becomes calculator.add
	app.WithHandler("add", func(ctx context.Context, payload []byte) (interface{}, error) {
		// payload is the entire request data
		// get full request if needed
		msg := rmb.GetRequest(ctx)
		if msg.Schema != rmb.DefaultSchema {
			return nil, fmt.Errorf("expecting schema to be %s", rmb.DefaultSchema)
		}

		var numbers []float64

		if err := json.Unmarshal(payload, &numbers); err != nil {
			return nil, fmt.Errorf("failed to load request payload was expecting list of float: %w", err)
		}

		var result float64
		for _, v := range numbers {
			result += v
		}

		return result, nil
	})

	// this function then becomes calculator.sub
	app.WithHandler("sub", func(ctx context.Context, payload []byte) (interface{}, error) {
		// implement sub here
		return nil, nil
	})

	peer := peer.NewPeer(..., router.serve);

        select{}
```

the `router.serve` being the single entry point for the router, that accepts the requests, routes them, handle answers and send them back to callers. 

> you might change the signature of the direct client handler function. I think you will need a reference to self (the direct client) to be able to send back responses

muhamadazmy:
### Description
Some refactoring to the low level Peer to be able to build the Router handler 

### Changes
- Clean up and clearer Peer interface

### Related Issues
- Fixes #499

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


rmb-sdk-go/peer/peer.go
```diff
@@ -367,15 +406,37 @@ func (d *Peer) Send(ctx context.Context, id string, twin uint32, fn string, data
...
 
-	if err := d.request(ctx, request); err != nil {
+	if err := d.send(ctx, request); err != nil {
 		return err
 	}
 
 	return nil
+}
+
+func (d *Peer) ParseResponse(response *types.Envelope, callBackErr error) ([]byte, error) {
```
muhamadazmy: This does not belong on the Peer at all.

- Interpreting the meaning of the envelope payload is up to the app that is using the peer (int he call back)
- Who said that the schema Always has to be `json`, what if i am using a binary payload of some sort?
- I don't see the `*Peer` object (d) being used, it means this function does not have to be on the Peer, instead can be defined as a stand alone utility function 

i suggest you take this out to be a stand alone function and rename it (since the function does not do any parsing instead it just extracts the payload and validate the schema

This function instead should be 

```go
// Json validates that the envelope holds json data and return the plain payload
func Json(envelope, err) ([]byte, error) {
}
```

rmb-sdk-go/examples/peer_server/peer/main.go
```diff
@@ -0,0 +1,74 @@
+package main
```
muhamadazmy: we already have this exact example under `peer/examples' why you added this one?

rmb-sdk-go/examples/peer_server/server/main.go
```diff
@@ -0,0 +1,83 @@
+package main
```
muhamadazmy: this should be moved to `peer/example/server` why did you place it here?

rmb-sdk-go/peer/peer.go
muhamadazmy: Another thought, may be since the Peer always do "json" on sending messages with no choice of encoding or schema we have 2 options:
- Always use json, in that case the schema validation and extraction of the error can be done before calling the handler callback. (the handler still need to access the data from the envelope) 
- Make the peer completely transparent to encoding/decoding format and schema which means the user has to specify this as a value during creating the peer.
  - This will make the NewPeer arguments takes a lot of arguments, may be it's better if we instead create a PeerOptions object to deal with arguments and default values.

rmb-sdk-go/peer/router.go
```diff
@@ -0,0 +1,133 @@
...
+		return
+	}
+
+	payload := env.Payload.(*types.Envelope_Plain).Plain
+	cmd := env.GetRequest().Command
+
+	response, err := r.call(ctx, cmd, payload)
+
+	// send response
+	if err := peer.SendResponse(ctx, uuid.NewString(), env.Source.Twin, env.Source.Connection, err, response); err != nil {
```
muhamadazmy: response HAS to use the same ID for request, not a new one. Obviously you never tested if your code works. Using a new ID means the caller won't be able to map the response to the request he sent.

rmb-sdk-go/peer/router.go
```diff
@@ -0,0 +1,133 @@
...
+
+	r.handlers[subCommand] = handler
+}
+
+// Use adds a middleware to the router
+func (r *Router) Use(mw Middleware) {
+	r.mw = append(r.mw, mw)
+}
+
+func (r *Router) Serve(ctx context.Context, peer Peer, env *types.Envelope, err error) {
```
muhamadazmy, line 73: There are some issues with implementation:
- The handler has no way to get the "twin" who sent this message. the Router in the other implementation where u copied most of the code also injects extra values for handler (per request) that holds some request information that can be fetched (sender twin id, and even the full envelope) in case it needs information from it.
- The router has NO way to handle requests in parallel. All handling you do is completely sequential (receive a message, handle, it, send response. Means the peer can't call your handler again until you return which make it sequential.

muhamadazmy: Instead when u receive a message you need to first, fill in the ctx with the extra data for the handler. then immediately start a go routine to process the message, and send response back

muhamadazmy (commented):
Please apply requested changes

rmb-sdk-go/peer/peer.go
rawdaGastan: opened an issue for the feature
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/507

rawdaGastan (approved):

Issue #484, opened 2023-11-14
rewrite the farmerbot in go


xmonader:
there's a lot of instability in the v version combined with the instability in components + the requirement of having rmb-peer and http server exposing the sdk-ts functionality.

Rawda did some initial work on the a farmerbot version before in go, and it's time to revive, and revise it. Please also introduce simplification in the configurations format, toml or yaml should be very intuitive options ✌🏼

rawdaGastan:
v repo: https://github.com/threefoldtech/farmerbot
go repo: https://github.com/rawdaGastan/farmerbot

farmerbot:

we have 3 managers:
1. node manager
2. farm manager
3. power manager

Jobs can be sent to the farmerbot using RMB

farm manager jobs:
1. define: to define a new farm

node manager jobs:
1. define: to define a new node
3. fine_node: to find a node (returns a node ID)
```
- _has_gpus_ => if you require one or more gpus you can filter on that with this parameter (should be a positive value)
- _gpu_vendors_ => a list of strings that will be used to filter the nodes on gpu vendor (for example AMD)
- _gpu_devices_ => a list of strings that will be used to filter the nodes on gpu device (for example GTX 1080)
- _certified_ => whether or not you want a certified node (not adding this argument means you don't care whether you get a certified or non certified node)
- _public_config_ => whether or not you want a node with a public config (not adding this argument means you don't care whether or not the node has a public config)
- _public_ips_ => how much public ips you need
- _dedicated_ => whether you want a dedicated node (rent the full node)
- _node_exclude_ => the node ids you want to exclude in your search
- _required_hru_ => the amount of hru required in kilobytes (add suffix mb or gb to define the required resources in megabyte or gygabite)
- _required_sru_ => the amount of sru required in kilobytes (add suffix mb or gb to define the required resources in megabyte or gygabite)
- _required_mru_ => the amount of mru required in kilobytes (add suffix mb or gb to define the required resources in megabyte or gygabite)
- _required_cru_ => the amount of cru required
```
power manager jobs:
1. power on: powers on a node
2. power off: powers off a node
3. configure: configures power
```
- wake_up_threshold => a value between 50 and 80 defining the threshold at which nodes will be powered on or off. If the usage percentage (total used resources devided by the total amount of resources) is greater then the threshold a new node will be powered on. In the other case the farmerbot will try to power off nodes if possible.
[...]

PR #509, opened 2023-11-22
add farmerbot

rawdaGastan:
### Related issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/484


farmerbot/cmd/farmerbot.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		log.Fatal().Err(err).Send()
+		os.Exit(1)
+	}
+}
+
+func init() {
+	zerolog.SetGlobalLevel(zerolog.InfoLevel)
+	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
+
```
xmonader: I believe we should support loading from env as well

farmerbot/slice/contains.go
```diff
@@ -0,0 +1,11 @@
+package slice
```
xmonader: can now use https://pkg.go.dev/slices@master#Contains

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,293 @@
...
+		identity:     identity,
+	}, nil
+}
+
+func (f *FarmerBot) start(ctx context.Context) error {
+	return f.powerManager.PowerOnAllNodes()
+}
+
+func (f *FarmerBot) stop(ctx context.Context) error {
+	return f.powerManager.PowerOnAllNodes()
```
xmonader: I'm not sure I get this, why stop powers on All nodes?

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,293 @@
...
+		log.Fatal().Err(err).Msg("[FARMERBOT] failed to create farmerbot direct peer")
+	}
+
+	select {}
+}
+
+func (f *FarmerBot) authorize(ctx context.Context, payload []byte) (context.Context, error) {
+	twinID := peer.GetTwinID(ctx)
+	if twinID != f.twinID {
+		return ctx, fmt.Errorf("you are not authorized for this action")
```
xmonader: tell the twin information and the expected one please

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,293 @@
...
+	if err != nil {
+		return err
+	}
+
+	balance, err := sub.GetBalance(accountAddress)
+	if err != nil && !errors.Is(err, substrate.ErrAccountNotFound) {
+		return errors.Wrap(err, "failed to get a valid account")
+	}
+
+	if balance.Free.Cmp(big.NewInt(20000)) == -1 {
```
xmonader: is that 20000 TFT or mill TFT?

farmerbot/models/config.go
```diff
@@ -0,0 +1,226 @@
...
+	ExcludedNodes []uint32 `json:"excluded_nodes" yaml:"excluded_nodes" toml:"excluded_nodes"`
+	Power         Power    `json:"power" yaml:"power" toml:"power"`
+}
+
+// Config is the configuration data for farmerbot
+type Config struct {
+	Farm  substrate.Farm `json:"farm" yaml:"farm" toml:"farm"`
+	Nodes []Node         `json:"nodes" yaml:"nodes" toml:"nodes"`
+	Power Power          `json:"power" yaml:"power" toml:"power"`
+	*sync.Mutex
```
xmonader: please make it `m sync.Mutex`, embedding mutex is dangerou as anyone with access to the struct can do mystruct.Lock, .. etc

xmonader (changes requested):

farmerbot/server/farmerbot.go
rawdaGastan: mill TFT

farmerbot/cmd/farmerbot.go
```diff
@@ -0,0 +1,120 @@
...
+	network, err = cmd.Flags().GetString("network")
+	if err != nil {
+		err = errors.Wrapf(err, "invalid network input '%s'", network)
+		return
+	}
+
+	if network != "dev" && network != "qa" && network != "test" && network != "main" {
+		err = fmt.Errorf("network must be one of dev, qa, test, and main not '%s'", network)
+		return
+	}
```
muhamadazmy: I would prefer network names to be defined as constants that are then used later on by the code.

Also a check like that would be better more readable, specially if the list of possible network is also defined as constant

```suggestion
	if !slices.Contains([]string{"dev", "qa", "test", "main"}, network) {
		err = fmt.Errorf("network must be one of dev, qa, test, and main not '%s'", network)
		return
	}
```

farmerbot/cmd/farmerbot.go
```diff
@@ -0,0 +1,120 @@
...
+
+	err := farmerBotCmd.Execute()
+	if err != nil {
+		log.Fatal().Err(err).Send()
+	}
+}
+
+func init() {
+	farmerBotCmd.Flags().StringP("config", "c", "", "enter your config file that includes your farm, node and power configs. Available formats are [json, yml, toml]")
+	farmerBotCmd.Flags().StringP("network", "n", "dev", "the grid network to use")
```
muhamadazmy: the default network should always be `main`

farmerbot/cmd/farmerbot.go
```diff
@@ -0,0 +1,120 @@
...
+	farmerBotCmd.Flags().StringP("mnemonic", "m", "", "the mnemonic of the account of the farmer")
+	farmerBotCmd.Flags().StringP("log", "l", "farmerbot.log", "enter your log file path to debug")
```
muhamadazmy: Modern apps don't use log files, (unless there are multiple log streams) normally an app need to log normally to console, it's up to the process manager (systemd, or zinit) to capture the logs and write it to the proper location (or not)

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,296 @@
...
+
+	farmerbot.dataManager = dataManager
+
+	return farmerbot, nil
+}
+
+// Run runs farmerbot to update nodes and power management
+func (f *FarmerBot) Run(ctx context.Context) {
+	if err := f.start(ctx); err != nil {
+		log.Fatal().Err(err).Msg("[FARMERBOT] error starting")
```
muhamadazmy: There are multiple points at this line:
- Why calling `Fatal()` here instead of actually just returing an erro ? It's not up to you to decide to terminate the application instead it's the caller (user) of the FarmerBot object to decide if it wants to stop the process or not, what if I wanna do something before i terminate in error. In general it's not acceptable that a library code to terminate the application instead of returning an error.
- Second comment on this is how you prefix every single log line with `[FARMERBOT]` !
  - Frist, why not use the zerolog fields, like `log.Debug().Str("module", "farmerbot").Msg()` 
  - This of course is also very verbose, instead you can do 
```go
import (
    zlog "github.com/rs/zerolog/log" 
)
// define a global variable in the farmerbot package like
var (
   log = zlog.With().Str("module", "farmerbot").Logger()
)

// then in function code
func something() {
   log.Debug().Msg("error starting")
}
```

One last comment:
- If your function needs to return an error, you should never log that error, instead you bubble the error up may be with added context like `return fmt.Errorf("failed to start: %w", err)` **but** never log that.
- It's up to the caller to LOG if the error won't bubble up anymore

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,296 @@
...
+	err = f.config.Set(&f.sub, inputs)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (f *FarmerBot) start(ctx context.Context) error {
+	return f.powerManager.PowerOnAllNodes()
```
muhamadazmy: I find it weird that this happens on start, there is absolutely no reason why I need to power on all nodes when i start, the farmerbot can first do what it needs to do until it comes to a possible power configuration then bring it to reality by powering on/off nodes.

On start, a node that has been off for a long time might need to be power cycled to make sure it reports uptime

farmerbot/server/farmerbot.go
muhamadazmy: @xmonader I think the reason is the farmer might be taking the farmerbot completely down which means it needs to leave the farm in good (up) state. Since if some nodes are down they can be stay down forever with no (easy) way to bring them up again.

Although i find this "extreme" also, what if the farmerbot is just being updated, it would be shame to power up all nodes just for that.

Hence i think it would be better if the farmerbot has other utilities subcommands like
```go 
// run the farmerbot in normal usage
bot run 
// start a single node given an id, or `all`
bot start <node-id> or `all`
// stop a single node given an id or `all`
bot stop <node-id> or `all`
```

Then if a farmer wants to opt out of using the bot, he can simply shut it down, then run the `bot start all` command

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,296 @@
...
+	powerManager *manager.PowerManager
+	dataManager  manager.DataManager
+	twinID       uint32
+	network      string
+	mnemonic     string
+	identity     substrate.Identity
+}
+
+// NewFarmerBot generates a new farmer bot
+func NewFarmerBot(ctx context.Context, configFile, network, mnemonic string) (FarmerBot, error) {
```
muhamadazmy: I think i explained before why using a `configFile` is wrong and that this instead need to be a `BotOptions` or whatever name `struct`.

The bot being a `library` code you should not assume the user providing the config in a specific way (here a file) instead it's up to the user to constrct the proper config either statically in code, from a file, env vars, etc...

Hence while the command line is okay to accept config file, the NewFarmerBot need to receive a concrete config object of sometype.

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,296 @@
...
+	farmerbot := FarmerBot{mnemonic: mnemonic, network: network, configFile: configFile}
+
+	substrateManager := substrate.NewManager(constants.SubstrateURLs[network]...)
+	sub, err := substrateManager.Substrate()
```
muhamadazmy: creating a single connection and holding it during the entire live of the bot is wrong. Instead, when a connection is need, the manager (that need to be kept ont he FarmerBot, not the conneciton) is used to get a connection as follows:

```go
cl, err := b.manager.Substrate()
// check error
defer cl.Close()
```

This way you sure that the manager will actually cycle through the provided substrate urls. Also connection are granted to be active.

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,296 @@
...
+	if err != nil {
+		return FarmerBot{}, err
+	}
+
+	farmerbot.sub = *sub
```
muhamadazmy: Why not first create the `manager` then create the farmerbot 
```
sub = connection()
bot = FarmerBot{sub: sub}
```
Also i don't get why the dereferencing here. I mean why not instead have the type as
```go
type FarmerBot stuct{ 
    sub *substrate.Substrate
}
```

> I still think this should be the full manager not a single connection the manager then can fallback to create connections against the other urls providing while creating

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,296 @@
...
+		return FarmerBot{}, err
+	}
+
+	// TODO: check farm hex seed if available
+	identity, err := substrate.NewIdentityFromSr25519Phrase(mnemonic)
+	if err != nil {
+		return FarmerBot{}, err
+	}
+
+	farmerbot.identity = identity
```
muhamadazmy: I still find this weird, why not FIRST construct all the bot attributes THEN create the bot, isntead of setting the attributes one by one.
```
identity = create()
manger = create()
twinid = create()

Bot {
   identityt,
   manager,
   twinid
}
```

farmerbot/manager/rmb.go
```diff
@@ -0,0 +1,67 @@
...
+	GetStoragePools(ctx context.Context, nodeTwin uint32) (pools []pkg.PoolMetrics, err error)
+	ListGPUs(ctx context.Context, nodeTwin uint32) (gpus []models.GPU, err error)
+}
+
+type rmbNodeClient struct {
+	rmb        rmb.Client
+	rmbTimeout time.Duration
+}
+
+func NewRmbNodeClient(rmb rmb.Client) rmbNodeClient {
```
muhamadazmy: Public method return a private object.  
```suggestion
func NewRmbNodeClient(rmb rmb.Client) *RmbNodeClient {
```

> See how i also returned a pointer not a value, because it's `*RmbNodeClient` what implementes RMB not `RmbNodeClient`

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,296 @@
...
+		&f.sub,
+		false,
+		router.Serve,
+	)
+
+	if err != nil {
+		log.Fatal().Err(err).Msg("[FARMERBOT] failed to create farmerbot direct peer")
+	}
+
+	select {}
```
muhamadazmy:
```suggestion
	<-ctx.Done()
```

we use `select{}` when we want to block forever with no chance of unlocking, but we already have a context that can be cancelled then `<-ctx.Done()` makes more sense , right ?

**not right**. Because this method (in this specific use case) does not need to be blocked, in the matter of fact you don't even need to start it in a `go routine`. I mean think about it, why start it in the background then block it from returing.

Creating the Peer makes it start serving requests immedilatey already in the background. Which means this method itself can return (also you can return an error)

 I used `select{}` in the example code because there was only the main routing and there was nothing that can stop the application from exiting completely, unlike here. So it's important to understand why things are done and change it to work with our usecase.

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,296 @@
...
+		f.mnemonic,
+		constants.RelayURLS[f.network],
+		fmt.Sprintf("farmerbot-%d", f.config.Farm.ID),
+		&f.sub,
+		false,
+		router.Serve,
+	)
+
+	if err != nil {
+		log.Fatal().Err(err).Msg("[FARMERBOT] failed to create farmerbot direct peer")
```
muhamadazmy: Please read the following comment, then instead return an error as explaing in various previous comments

farmerbot/parser/parser.go
```diff
@@ -0,0 +1,55 @@
...
+
+	return content, filepath.Ext(path)[1:], nil
+}
+
+// ParseIntoInputConfig parses the configuration
+func ParseIntoInputConfig(content []byte, format string) (models.InputConfig, error) {
+	input := models.InputConfig{}
+
+	var err error
+	switch {
```
muhamadazmy: Who decided that we need multiple file formats? Just a question. I find it over engineering tbh

farmerbot/models/config.go
```diff
@@ -0,0 +1,231 @@
...
+// Config is the configuration data for farmerbot
+type Config struct {
+	Farm  substrate.Farm `json:"farm" yaml:"farm" toml:"farm"`
+	Nodes []Node         `json:"nodes" yaml:"nodes" toml:"nodes"`
+	Power Power          `json:"power" yaml:"power" toml:"power"`
+	m     sync.Mutex
+}
+
+// Set sets the config data from configuration inputs
+func (c *Config) Set(sub Sub, inputs InputConfig) error {
```
muhamadazmy: I find this entire function improper, since this is a configuration model it's pure data that provide input to the place where it handles the logic (farmerbot)

In other words, this is not the right place to have a business logic. don't you agree?

In ohter words.
- load config from source (file, static, etc..)
- pass config as is to `module`
- module interpret the config to build it's own state

farmerbot/models/config.go
```diff
@@ -0,0 +1,231 @@
...
+			c.Nodes = append(c.Nodes, configNode)
+		}
+	}
+
+	return nil
+}
+
+// GetNodeByNodeID gets a node by id
+func (c *Config) GetNodeByNodeID(nodeID uint32) (Node, error) {
+	for _, n := range c.Nodes {
```
muhamadazmy: Why not store the nodes of the farm in a map instead?

farmerbot/models/config.go
```diff
@@ -0,0 +1,231 @@
...
+	err := c.SetConfigNodes(sub, inputs)
+	if err != nil {
+		return err
+	}
+
+	return c.validate()
+}
+
+// SetConfigNodes sets the config nodes from configuration inputs
+func (c *Config) SetConfigNodes(sub Sub, i InputConfig) error {
```
muhamadazmy: Again while this is the wrong place, I also noticed that this only is done when the bot is starting. Which is wrong, instead the list of the nodes need (ids) need to be synced with the chain with every bot iteration.

farmerbot/models/config.go
```diff
@@ -0,0 +1,231 @@
...
+
+	return Node{}, fmt.Errorf("node %d not found", nodeID)
+}
+
+// UpdateNode updates a node in the config
+func (c *Config) UpdateNode(node Node) error {
+	c.m.Lock()
+	defer c.m.Unlock()
+
+	for i, n := range c.Nodes {
```
muhamadazmy: use a map please.

farmerbot/models/config.go
muhamadazmy: In other words, we need to make sure that the node ids the bot knows about must be exactly the same as the nodes (matching including/excluding lists) on the grid.

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,296 @@
...
+
+	log.Info().Msg("[FARMERBOT] up and running...")
+
+	// graceful shutdown
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+	<-sigChan
+
+	shutdownCtx, shutdownRelease := context.WithTimeout(context.Background(), 10*time.Second)
+	defer shutdownRelease()
```
muhamadazmy: It's not right to do, Why the farmer bot care how or when the context would be cancelled.  There is a HUGE different between library and cmdline application.

The farmer bot takes a context, all what it needs to do is graceful shutdown (basically return from the main loop) if the ctx was cancelled it shoud not care how the context is canelled.

The user of the FarmerBot object then can decide when and how the context will be cancelled. The command line can then listen on signals, os something else, and cancel the context accordingly

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,296 @@
...
+	return farmerbot, nil
+}
+
+// Run runs farmerbot to update nodes and power management
+func (f *FarmerBot) Run(ctx context.Context) {
+	if err := f.start(ctx); err != nil {
+		log.Fatal().Err(err).Msg("[FARMERBOT] error starting")
+	}
+
+	go f.update(ctx)
```
muhamadazmy: Since update is the main farmerbot loop, why do u run it in routing then block on a signal? instead i will do something like this 
```
for {
   select {
        ctx.Done(): return
        <-sleep(until next rune):
        // do your job
}
```

farmerbot/server/farmerbot.go
```diff
@@ -0,0 +1,296 @@
...
+
+func (f *FarmerBot) start(ctx context.Context) error {
+	return f.powerManager.PowerOnAllNodes()
+}
+
+func (f *FarmerBot) stop(ctx context.Context) error {
+	return f.powerManager.PowerOnAllNodes()
+}
+
+func (f *FarmerBot) update(ctx context.Context) {
```
muhamadazmy: I can't explicelty state everything wrong in this approach but it's a general feeling of un-necessary compexity. The only concrete statemeng i can give is that:

Instead of doing everything in hidden subcalls that iterate over the entire shared state (this makes it really really hard to follow logic or debug) we should instead process each node in a single iteration

```
get list of ndoes from that farm that matches our config.
usedCap = 0
totalCap = 0
for node in nodes {
  - do we have a node state associdated with this node id. otherwise, fetch!
  - if it's off, do we need to wake it up for random wake ups? 
  - if node is on incrememnt both usedCap and totalCap
}
check percentage of usedCap to totalCap and then take action accordingly
```

this way the logic is straight forward and can be followed and debugged easily. 

> **NOTE** the problem as i see it is that you trying to follow the same design as the old one (with separate manager and so) which might not be necessary.

muhamadazmy: Another thing that i find difficult about this function is you can't undesrand what it does by reading the function name, and logic.
for example what does f.dataManager.Update() means? or f.powerManager.PowerManager() what does that do. 

Hence the logic imho need to be cleaned and written as above, since it makes it easier to follow. It's not nice that u do the datamanager.Update() then do powermanager.Whatever() which has access to the shared state of the data manger.

I think splitting this into different manager make it harder and not improving abstration.

muhamadazmy (changes requested):
Please take a look at my comments, and redisgn according. There is nothing that forces us from using (again) the same paradigm used in the V version. 

If this is still hard to get right; we can do a peer programing session on Monday, to settle on the design then leave you to the details.

This is the first iteration of the review, i want to get the main lines correct first before we get into the smaller stuff

farmerbot/internal/farmerbot.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	log.Info().Msg("up and running...")
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			startTime := time.Now()
```
muhamadazmy: There is a problem here let me walk you through it to explain
- The loop starts, and immediately does the select.
- since the ctx is __not__ cancelled, the select fall immediately into `default` branch
- the logic executes then sleeps for `timeStoSleep` minutes.
- during the sleep time, you have **NO** way to actually detect the ctx cancellation (no select or wait on the ctx done channel)
- this means if the ctx is cancelled u won't detect that (and return) until the sleep time is over.

This can be fixed like
```suggestion
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-time.After(sleepTime):
		     // logic
```

This means that the runtime will wait on both the ctx cancellation and the sleep at the same time, executing the branch whenver one is done first.

But this introduce another problem that u probably can already see, which is that the wait happens before running the logic the first time!

but this can be fixed easily as follows:

```go
for {
          /// LOGIC happens first

         // at the end of the loop
         select {
               case <-ctx.Done():
                	return nil
               case <-time.After(sleepTime):
         }
}
```

This means when the loop starts, the logic is excuted first, then both ctx cancellation and the wait are polled on at the same time and execute whatever happens first. (being that the ctx cancellation returns, and the sleep is over to just continue

muhamadazmy: I would also split this method because it's so indented  as follows
```
for {
     f.iterate() 

     select {
            ...
     }
} 
``` 

this means `iterate()` can create and close the substrate connection instead of keeping it open all the time like in this implementation

farmerbot/internal/farmerbot.go
```diff
@@ -0,0 +1,319 @@
...
+
+			log.Debug().Msg("check nodes latest updates")
+			for _, node := range f.state.nodes {
```
muhamadazmy: We agreed that the list of the nodes need to be synched with the chain on every iteration of the main loop this means:

Before you do `for node in f.nodes` you need to make the following:
- Fetch list of ALL farm node ids
- If an id is in the <node ids list> but not in your map, then fetch the node and update the state for that node
- If your state map has a node that does not exist in the id list it needs to be dropped.

farmerbot/internal/farmerbot.go
```diff
@@ -0,0 +1,319 @@
...
+				neverShutDown := slices.Contains(f.state.config.NeverShutDownNodes, uint32(node.ID))
+				hasClaimedResources := node.timeoutClaimedResources.After(time.Now())
+				dedicatedFarm := f.state.farm.DedicatedFarm
```
muhamadazmy: I don't get the benefit if having a `state` object on the farm, why this is just not embedded the farmerbot object itself ? I mean why do i have to do `f.state.farm` instead of `f.farm` ?

farmerbot/internal/farmerbot.go
muhamadazmy: This means that you don't need to create the noes map on start the way u do it but this can happen naturally in your iteration.
```
nodeIDs = farm.GetNodes() | filter based on config (include/exclude)
if map contains ids that are not in nodeIDs {
   drop(node from map)
}

for id in nodeIDs {
    if node not in map of nodes {
        collect node information, and insert in map
    }

    // management code
}
```

I really thought this was clear from our previous discussions

farmerbot/internal/farmerbot.go
muhamadazmy: if `dedicatedFarm` or `overProvisionCPU` are global configuration, why do u need to store this on every node?

farmerbot/internal/types.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/threefoldtech/zos/pkg"
+	"github.com/threefoldtech/zos/pkg/gridtypes"
+)
+
+// Config is the inputs for configuration for farmerbot
+type Config struct {
+	FarmID             uint32   `yaml:"farm_id"`
+	IncludedNodes      []uint32 `yaml:"included_nodes"`
+	ExcludedNodes      []uint32 `yaml:"excluded_nodes"`
+	NeverShutDownNodes []uint32 `yaml:"never_shutdown_nodes"`
```
muhamadazmy, line 20: What is the difference between this and the `ExcludeNodes` ?

farmerbot/internal/farmerbot.go
```diff
@@ -0,0 +1,319 @@
...
+
+				ltsNode, err := getNodeWithLatestChanges(ctx, subConn, f.rmbNodeClient, uint32(node.ID), neverShutDown, hasClaimedResources, dedicatedFarm)
```
muhamadazmy: confusing function name, after reading the logic I was thinking another name like `getNode` or something because it literary just gets an actual node state from real data

farmerbot/internal/power_manager.go
```diff
@@ -0,0 +1,399 @@
...
+
+	offNodes := p.state.filterNodesPower([]powerState{off})
+
+	periodicWakeUpStart := p.state.config.Power.PeriodicWakeUpStart.PeriodicWakeUpTime()
+
+	var wakeUpCalls uint8
+	nodesLen := len(p.state.nodes)
+
+	for _, node := range offNodes {
+		// TODO: why??
```
muhamadazmy: Why what?

farmerbot/internal/power_manager.go
```diff
@@ -0,0 +1,399 @@
...
+		if now.Day() == 1 && now.Hour() == 1 && now.Minute() >= 0 && now.Minute() < 5 {
```
muhamadazmy: I honestly don't get this logic? what are u trying to do ?

farmerbot/internal/power_manager.go
```diff
@@ -0,0 +1,399 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+	substrate "github.com/threefoldtech/tfchain/clients/tfchain-client-go"
+	zos "github.com/threefoldtech/zos/client"
+)
+
+// PowerManager manages the power of nodes
+type PowerManager struct {
+	state    *state
```
muhamadazmy: Enough with the shared state please. That's exactly what we agreed upon and also during scrum this morning. if u gonna share state between multiple objects that just live internally inside the same package i would rather then have them one object.

why not call  `powerManger.PeriodicWakeup` with some arguments instead so i don't have to share state between objects. 

In the matter of fact why do I need a power manager if it's just a function that based on some calculation it can decide to (or not) wake the up the node 

I explained in a previous PR that the logic should instead be straight forward like
```
 for nodes in farm nodes {
      nodeData = getNode() // fetches state or whatever 

        f.shouldWakeUp(&node) // this one basically checks if the node need to be nudged based on time stamps and so
}
```

farmerbot/internal/power_manager.go
```diff
@@ -0,0 +1,399 @@
...
+	}
+
+	node.powerState = shuttingDown
+	node.lastTimePowerStateChanged = time.Now()
+
+	return p.state.updateNode(node)
+}
+
+// FindNode finds an available node in the farm
+func (p *PowerManager) FindNode(sub Sub, nodeOptions NodeOptions) (uint32, error) {
```
muhamadazmy: It doesn't feel like a `FindNode` belongs to something called `power manager`

Note: I already stated multiple times that I don't like that we have that those called managers which all share the same state, so basically they can become small **parameterised** methods on the main farmerbot object.

farmerbot/internal/power_manager.go
```diff
@@ -0,0 +1,399 @@
...
+	now := time.Now()
+
+	offNodes := p.state.filterNodesPower([]powerState{off})
+
+	periodicWakeUpStart := p.state.config.Power.PeriodicWakeUpStart.PeriodicWakeUpTime()
+
+	var wakeUpCalls uint8
+	nodesLen := len(p.state.nodes)
+
+	for _, node := range offNodes {
```
muhamadazmy: as i already mentioned in the previous review please this can done on `node by node` bases in the main loop. for example why this instead doesn't become 
```
 /// THIS IS MAIN LOOP NOT THIS LOOP, THE ONE IN farmertbo.Run
  for node in nodes {
         node = getLatestState() 
          f.periodicWakeUp(&node) // this one decides if node to be powerd on or not
   } 
``

File: farmerbot/internal/state.go [hunk 37]
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	configNode.hasActiveRentContract = rentContract != 0
+
+	powerTarget, err := sub.GetPowerTarget(nodeID)
+	if err != nil {
+		return node{}, errors.Wrapf(err, "failed to get node %d power target from substrate", nodeID)
+	}
+
+	if powerTarget.State.IsUp && powerTarget.Target.IsUp && configNode.powerState != on {
```
muhamadazmy, line 131: how the `configNode.powerState` will has any value if you just created configNode and never assigned anything to powerState

And since the zero value of powerState is ON it means this condition will never ever met

muhamadazmy (changes requested):
I feel this is still too complex and not clear as it should. Most of my comments about architecture which i still think is far from optimal. And i feel like i need to re add the same comments from last PR.

In general please consider the following:
- Shared internal state is a no no. If you feel u need to share internal state then u are doing something wrong 
- again, state is private, no object need to expose his internal state, if u do so then you are doing something wrong.
- Use small and **parameterised** functions for example `PeriodicWakeUp(&node)` against `PeriodicWakeup()` the first one is better, because it checks this **specific** node if it can be waked up or not.
- Smaller functions, leads to clearer and smaller function names.

farmerbot/internal/types.go
rawdaGastan: `ExcludedNodes` are not managed (capacity planing and power management) by the farmerbot while `NeverShutDownNodes` are included in the capacity planing but not in the power management

xmonader:
It might be bit too much, but can we check the close issues (and the open ones) https://github.com/threefoldtech/farmerbot/issues (ofc we can ignore anything related to typescript or rmb-peer .. etc) to make sure things are already covered

farmerbot/parser/parser.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/threefoldtech/tfgrid-sdk-go/farmerbot/internal"
+	"gopkg.in/yaml.v3"
+)
+
+// ReadFile reads a file and returns its contents
+func ReadFile(path string) ([]byte, error) {
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return []byte{}, err
```
xmonader: could have been `nil, err` instead of `[]byte{}, err`

farmerbot/README.md
```diff
@@ -0,0 +1,207 @@
...
+  - "<your node ID to be included, required at least 2>"
+excluded_nodes:
+  - "<your node ID to be excluded, optional>"
+never_shutdown_nodes:
+  - "<your node ID to be never shutdown, optional>"
+power:
+  periodic_wake_up_start: "<daily time to wake up nodes for your farm, default is the time your run the command, format is 00:00AM or 00:00PM, optional>"
+  wake_up_threshold: "<the threshold number for resources usage that will need another node to be on, default is 80, optional>"
+  periodic_wake_up_limit: "<the number (limit) of nodes to be waken up everyday, default is 1, optional>"
+  overprovision_cpu: "<how much node allows over provisioning the CPU , default is 1, range: [1;4], optional>"
```
xmonader: TBH, I don't get what this does and what the number change affects is it like a multiplier or extra N cores

farmerbot/README.md
```diff
@@ -0,0 +1,207 @@
...
+Farmerbot is a service that a farmer can run allowing him to automatically manage power of the nodes of his farm.
+
+## How to use
+
+- Create a new yml/yaml file `config.yml` and add your configurations:
+
+```yml
+farm_id: "<your farm ID, required>"
+included_nodes:
+  - "<your node ID to be included, required at least 2>"
```
xmonader: its IDs in plural

farmerbot/README.md
```diff
@@ -0,0 +1,207 @@
...
+  periodic_wake_up_limit: "<the number (limit) of nodes to be waken up everyday, default is 1, optional>"
+  overprovision_cpu: "<how much node allows over provisioning the CPU , default is 1, range: [1;4], optional>"
+```
+
+- [Download](#download) farmerbot binaries.
+
+- Run the bot
+
+```bash
+farmerbot run -c config.json -m <mnemonic> -n dev -d
```
xmonader: while the configuration in the section above is yaml, the instructions are to use json?

farmerbot/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Build
+
+Run the following command inside the directory:
+
+```bash
+make build
+```
+
+## Supported commands
```
xmonader, line 105: I'm wondering from usability perspective, how can I still use it, start, and start all and version commands don't seem to help or these are extra ones other than the default aforementioned

farmerbot/parser/parser.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	return content, nil
+}
+
+// ParseIntoConfig parses the configuration
+func ParseIntoConfig(content []byte) (internal.Config, error) {
```
xmonader, line 27: I think validation of the configurations needs to get expanded beyond the checking of overlapping nodes ranges
non exhaustive list based on the configurations mentioned in the README.md

- invalid farm ID
- invalid nodes IDs within that farm
- duplicate nodes 
- required parameters e.g in this case would be farmID on the yaml struct.
- overprovision in range 1-4
- invalid datetime format
- invalid threshold values e.g negative numbers, i think it should be within 1-100, it's utilizationWakeupThresholdPercent on all of the included nodes
- invalid network

farmerbot/internal/types.go
```diff
@@ -0,0 +1,194 @@
...
+	Power              power    `yaml:"power"`
+}
+
+type powerState uint8
+
+const (
+	on = powerState(iota)
+	wakingUP
+	off
```
xmonader, line 29: why shuttingDown in order is after off? or is that how it's coded on chain? because the order isn't intuitive

farmerbot/internal/types.go
```diff
@@ -0,0 +1,194 @@
...
+
+	powerState                powerState
+	timeoutClaimedResources   time.Time
+	lastTimePowerStateChanged time.Time
+	lastTimeAwake             time.Time
+	timesRandomWakeUps        int
+}
+
+// NodeOptions represents the options to find a node
+type NodeOptions struct {
```
xmonader: From the comment I thought it was NodeFilterOption correct? because `NodeExclude` isn't a NodeOption, but a NodeFilteringOption, and probably we should rename to `NodesExcluded`

farmerbot/internal/types.go
```diff
@@ -0,0 +1,194 @@
...
+
+// FreeCapacity calculates the free capacity of a node
+func (n *node) freeCapacity(overProvisionCPU float32) capacity {
+	total := n.resources.total
+	total.cru = uint64(math.Ceil(float64(total.cru) * float64(overProvisionCPU)))
+	return total.subtract(n.resources.used)
+}
+
+// ConsumableResources for node resources
+type consumableResources struct {
```
xmonader, line 115: what does consumableResources? how is it different from nodeResources? also I'm not sure if these are the consumed resources or the avaiable to consume?

farmerbot/internal/types.go
```diff
@@ -0,0 +1,194 @@
...
+// Subtract subtracts a new capacity
+func (cap *capacity) subtract(c capacity) (result capacity) {
+	result.cru = subtract(cap.cru, c.cru)
+	result.mru = subtract(cap.mru, c.mru)
+	result.sru = subtract(cap.sru, c.sru)
+	result.hru = subtract(cap.hru, c.hru)
+	return result
+}
+
+func subtract(a uint64, b uint64) uint64 {
```
xmonader: pleaes rename to `subtractOrZero`

farmerbot/internal/types.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	return 0
+}
+
+// Power represents power configuration
+type power struct {
+	WakeUpThreshold     uint8      `yaml:"wake_up_threshold"`
+	PeriodicWakeUpStart wakeUpDate `yaml:"periodic_wake_up_start"`
+	PeriodicWakeUpLimit uint8      `yaml:"periodic_wake_up_limit"`
+	OverProvisionCPU    float32    `yaml:"overprovision_cpu,omitempty"`
```
xmonader: why is this float32 while the range is 1-4 ?

farmerbot/internal/types.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// wakeUpDate is the date to wake up all nodes
+type wakeUpDate time.Time
+
+// UnmarshalText unmarshal the given TOML string into wakeUp date
+func (d *wakeUpDate) UnmarshalText(b []byte) error {
```
xmonader, line 177: the goal is to unquote the double quotes around a time string? `"3:04PM"` ? please rename it to `fromQuotedDateTime` and maybe we should return another value instead of the pointer semantics here

farmerbot/internal/types.go
```diff
@@ -0,0 +1,194 @@
...
+	s := strings.Trim(string(b), "\"")
+	t, err := time.Parse("03:04PM", s)
+	if err != nil {
+		return err
+	}
+	*d = wakeUpDate(t)
+	return nil
+}
+
+// MarshalText marshals the wake up TOML date
```
xmonader: rename to `toQuotedDateTime`

farmerbot/internal/substrate.go
```diff
@@ -0,0 +1,19 @@
+package internal
+
+import (
+	"github.com/centrifuge/go-substrate-rpc-client/v4/types"
+	substrate "github.com/threefoldtech/tfchain/clients/tfchain-client-go"
+)
+
+// Sub is substrate client interface
+type Sub interface {
```
xmonader: Can we use a more intutive name e.g Substrate, it seems like a Subbing interface to subsitute or to subtract :D

xmonader (changes requested):
Added a partial list of change requests, will be checking the mechanics tmw inshallah

farmerbot/README.md
```diff
@@ -0,0 +1,207 @@
+# Farmerbot
+
+<a href='https://github.com/jpoles1/gopherbadger' target='_blank'>![gopherbadger-tag-do-not-edit](https://img.shields.io/badge/Go%20Coverage-73%25-brightgreen.svg?longCache=true&style=flat)</a>
+
+Farmerbot is a service that a farmer can run allowing him to automatically manage power of the nodes of his farm.
```
mariobassem: them/their for gender inclusion :D

farmerbot/internal/constants.go
```diff
@@ -0,0 +1,57 @@
...
+	timeoutPowerStateChange = time.Minute * 30
+
+	// periodicWakeUpDuration is the duration for periodic wake ups
+	periodicWakeUpDuration = time.Minute * 30
+
+	defaultRandomWakeUpsAMonth = 10
+
+	// defaultPeriodicWakeUPLimit default number of nodes will be woken up every 5 minutes during a periodic wake up
+	defaultPeriodicWakeUPLimit = uint8(1)
+
```
xmonader: Why not use typed consts instead of the conversion?

farmerbot/internal/constants.go
```diff
@@ -0,0 +1,57 @@
...
+	// defaultCPUProvision number
+	defaultCPUProvision = float32(2)
```
xmonader: while this value is in range 1-4, why is it required to be a float?

farmerbot/internal/constants.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// defaultWakeUpThreshold default threshold to wake up a new node
+	defaultWakeUpThreshold = uint8(80)
+	// minWakeUpThreshold min threshold to wake up a new node
+	minWakeUpThreshold = uint8(50)
```
xmonader: I don't understand the relation of the WakeupThreshold parameters, The original idea was that I configure the farmerbot for a utilization percentage to start powering on nodes until the available online nodes resources summation is less than the wakeup threshold or there're no more nodes to turn no

farmerbot/internal/constants.go
```diff
@@ -0,0 +1,57 @@
...
+// SubstrateURLs for substrate urls
+var SubstrateURLs = map[string][]string{
+	TestNetwork: {"wss://tfchain.test.grid.tf/ws", "wss://tfchain.test.grid.tf:443"},
+	MainNetwork: {"wss://tfchain.grid.tf/ws", "wss://tfchain.grid.tf:443"},
+	DevNetwork:  {"wss://tfchain.dev.grid.tf/ws", "wss://tfchain.dev.grid.tf:443"},
+	QaNetwork:   {"wss://tfchain.qa.grid.tf/ws", "wss://tfchain.qa.grid.tf:443"},
+}
+
+// relayURLS relay urls
+var relayURLS = map[string]string{
```
xmonader: relayURLs like the above

farmerbot/cmd/root.go
```diff
@@ -0,0 +1,115 @@
...
+	startCmd.AddCommand(startAllCmd)
+
+	err := farmerBotCmd.Execute()
+	if err != nil {
+		log.Fatal().Err(err).Send()
+	}
+}
+
+func init() {
+	farmerBotCmd.PersistentFlags().StringP("network", "n", internal.MainNetwork, "the grid network to use")
```
xmonader: let's add a hint on the allowed networks and default it to mainnet

farmerbot/cmd/root.go
```diff
@@ -0,0 +1,115 @@
...
+	farmerBotCmd.PersistentFlags().StringP("mnemonic", "m", "", "the mnemonic of the account of the farmer")
+	farmerBotCmd.PersistentFlags().StringP("seed", "s", "", "the hex seed of the account of the farmer")
+	farmerBotCmd.PersistentFlags().BoolP("debug", "d", false, "by setting this flag the farmerbot will print debug logs too")
+	farmerBotCmd.MarkFlagsMutuallyExclusive("mnemonic", "seed")
+
+	runCmd.Flags().StringP("config", "c", "", "enter your config file that includes your farm, node and power configs. Available format is yml/yaml")
```
xmonader: Available -> Allowed and if there're file extension restriction please communicate it to the user

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+// state is the state data for farmerbot
+type state struct {
+	farm   substrate.Farm
+	nodes  map[uint32]node
+	config Config
+	m      sync.Mutex
+}
+
+// NewState creates new state from configs
+func newState(ctx context.Context, sub Sub, rmbNodeClient RMB, inputs Config) (*state, error) {
```
xmonader: inputs -> cfg

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+	// required from power for nodes
+	if state.config.Power.OverProvisionCPU == 0 {
+		state.config.Power.OverProvisionCPU = defaultCPUProvision
+	}
+
+	if state.config.Power.OverProvisionCPU < 1 || state.config.Power.OverProvisionCPU > 4 {
+		return nil, fmt.Errorf("cpu over provision should be a value between 1 and 4 not %v", state.config.Power.OverProvisionCPU)
+	}
+
+	// set farm
```
xmonader: is that comment for the setting farm on the state or by mistake?

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+
+	delete(s.nodes, nodeID)
+}
+
+// UpdateNode updates a node in the config
+func (s *state) updateNode(node node) error {
+	s.m.Lock()
+	defer s.m.Unlock()
+
+	_, ok := s.nodes[uint32(node.ID)]
```
xmonader: why not set it directly instead checking on existence? also notice the conversion of uint32(node.ID) maybe it's better to extract it in a variable (it's done 3 times)

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+	if !ok {
+		return fmt.Errorf("node %d is not found", uint32(node.ID))
+	}
+
+	s.nodes[uint32(node.ID)] = node
+
+	return nil
+}
+
+// FilterNodesPower filters ON or OFF nodes
```
xmonader: But the powerState is passed in as a slice, so not limited to ON and OFF states, Right?

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+			filtered[nodeID] = node
+		}
+	}
+	return filtered
+}
+
+// FilterAllowedNodesToShutDown filters nodes that are allowed to shut down
+//
+// nodes with public config can't be shutdown
+// Do not shutdown a node that just came up (give it some time)
```
xmonader: How long?

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+func (s *state) filterAllowedNodesToShutDown() map[uint32]node {
+	filtered := make(map[uint32]node, 0)
+	for nodeID, node := range s.nodes {
+		if node.isUnused() && !node.PublicConfig.HasValue &&
+			!node.neverShutDown && !node.hasActiveRentContract &&
```
xmonader: I suggest to create a method on the node e.g `canShutdown()`

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+			return fmt.Errorf("node %d: total CRU is required", n.ID)
+		}
+		if n.resources.total.mru == 0 {
+			return fmt.Errorf("node %d: total MRU is required", n.ID)
+		}
+		if n.resources.total.hru == 0 {
+			return fmt.Errorf("node %d: total HRU is required", n.ID)
+		}
+	}
+
```
xmonader, line 252: Please remember that wakeupThreshold is a percentage so between 1-100

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+	state.farm = *farm
+
+	nodes, err := fetchNodes(ctx, sub, rmbNodeClient, inputs, farm.DedicatedFarm)
+	if err != nil {
+		return nil, err
+	}
+
+	state.nodes = nodes
+
+	err = state.validate()
```
xmonader: can be `if err:=state.validate; err!=nil...`

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+type state struct {
+	farm   substrate.Farm
+	nodes  map[uint32]node
+	config Config
+	m      sync.Mutex
+}
+
+// NewState creates new state from configs
+func newState(ctx context.Context, sub Sub, rmbNodeClient RMB, inputs Config) (*state, error) {
+	state := state{config: inputs}
```
xmonader: please change to s, to avoid shadowing the state struct

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+		return nil, fmt.Errorf("cpu over provision should be a value between 1 and 4 not %v", state.config.Power.OverProvisionCPU)
+	}
+
+	// set farm
+	farm, err := sub.GetFarm(inputs.FarmID)
+	if err != nil {
+		return nil, err
+	}
+
+	state.farm = *farm
```
xmonader: we can move this setting down, when all of our validations done, so we can recover to a working state.

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+	sub Sub,
+	rmbNodeClient RMB,
+	nodeID uint32,
+	neverShutDown,
+	hasClaimedResources,
+	dedicatedFarm bool,
+	oldPowerState powerState,
+) (node, error) {
+
+	log.Debug().Uint32("ID", nodeID).Msg("Include node")
```
xmonader: remove debug line

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+	nodeObj, err := sub.GetNode(nodeID)
+	if err != nil {
+		return node{}, errors.Wrapf(err, "failed to get node %d from substrate", nodeID)
```
xmonader: I noticed the usage of fmt.Errorf and %w elsewhere, let's be consistent

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	configNode := node{
+		Node: *nodeObj,
+	}
+
+	configNode.neverShutDown = neverShutDown
```
xmonader: why didn't you set this in the configNode struct inlined in 105?

farmerbot/internal/state.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	s.nodes[uint32(node.ID)] = node
+
+	return nil
+}
+
+// FilterNodesPower filters ON or OFF nodes
+func (s *state) filterNodesPower(states []powerState) map[uint32]node {
+	filtered := make(map[uint32]node, 0)
```
xmonader: You don't need the map size hint, it's not guaranteed to be honored

farmerbot/internal/rmb.go
```diff
@@ -0,0 +1,56 @@
...
+
+// RMB is an rmb abstract client interface.
+type RMB interface {
+	Statistics(ctx context.Context, nodeTwin uint32) (stats zos.Counters, err error)
+	GetStoragePools(ctx context.Context, nodeTwin uint32) (pools []pkg.PoolMetrics, err error)
+	ListGPUs(ctx context.Context, nodeTwin uint32) (gpus []zos.GPU, err error)
+}
+
+type RMBNodeClient struct {
+	rmb        rmb.Client
```
xmonader: client instead of rmb

farmerbot/internal/types.go
```diff
@@ -0,0 +1,194 @@
...
+	ExcludedNodes      []uint32 `yaml:"excluded_nodes"`
+	NeverShutDownNodes []uint32 `yaml:"never_shutdown_nodes"`
+	Power              power    `yaml:"power"`
+}
+
+type powerState uint8
+
+const (
+	on = powerState(iota)
+	wakingUP
```
xmonader: I think it should be wakingUp

farmerbot/internal/power.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	node.powerState = wakingUP
+	node.lastTimeAwake = time.Now()
+	node.lastTimePowerStateChanged = time.Now()
+
+	return f.updateNode(node)
+}
+
+// PowerOff sets the node power state OFF
```
xmonader: - let's change docstring to powerOff
- include the node ID in error messages when possible please

farmerbot/internal/power.go
```diff
@@ -0,0 +1,191 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+)
+
+// PowerOn sets the node power state ON
```
xmonader: docstring: powerOn

farmerbot/internal/power.go
```diff
@@ -0,0 +1,191 @@
...
+		return errors.Wrapf(err, "failed to set node %d power target to down", nodeID)
+	}
+
+	node.powerState = shuttingDown
+	node.lastTimePowerStateChanged = time.Now()
+
+	return f.updateNode(node)
+}
+
+// manageNodesPower for power management nodes
```
xmonader, line 96: Not sure what does "manageNodesPower" do? does it bring nodes online?

farmerbot/internal/power.go
```diff
@@ -0,0 +1,191 @@
...
+func (f *FarmerBot) manageNodesPower(sub Sub) error {
+	nodes := f.filterNodesPower([]powerState{on, wakingUP})
+
+	usedResources, totalResources := calculateResourceUsage(nodes)
+	if totalResources == 0 {
+		return nil
+	}
+
+	resourceUsage := uint8(100 * usedResources / totalResources)
```
xmonader, line 105: Is it possible that totalResources could be a zero at any point? e.g broken zos reporting, invalid state, zero value that wasn't updated?

farmerbot/internal/power.go
```diff
@@ -0,0 +1,191 @@
...
+					log.Error().Err(err).Uint32("nodeID", uint32(node.ID)).Msg("Power off node")
+					nodesLeftOnline += 1
+					newUsedResources += node.resources.used.hru + node.resources.used.sru +
+						node.resources.used.mru + node.resources.used.cru
+					newTotalResources += node.resources.total.hru + node.resources.total.sru +
+						node.resources.total.mru + node.resources.total.cru
+				}
+			}
+		}
+	} else {
```
xmonader: make it 2 ifs do the <1 check first please

farmerbot/internal/power.go
```diff
@@ -0,0 +1,191 @@
...
+		log.Debug().Msg("Nothing to shutdown.")
+	}
+
+	return nil
+}
```
xmonader, line 203: This file can use some tests for the conditions I believe (from the first sight, everything checks out :)

farmerbot/internal/types.go
```diff
@@ -0,0 +1,194 @@
...
+	timeoutClaimedResources   time.Time
+	lastTimePowerStateChanged time.Time
+	lastTimeAwake             time.Time
+	timesRandomWakeUps        int
+}
+
+// NodeOptions represents the options to find a node
+type NodeOptions struct {
+	NodeExclude  []uint32 `json:"node_exclude,omitempty"`
+	HasGPUs      uint8    `json:"has_gpus,omitempty"`
```
xmonader: is that GPUs count? should be NumGPU instead of HasGPUs

farmerbot/internal/find_node.go
```diff
@@ -0,0 +1,154 @@
...
+	nodeOptionsCapacity := capacity{
+		hru: nodeOptions.HRU,
+		sru: nodeOptions.SRU,
+		cru: nodeOptions.CRU,
+		mru: nodeOptions.MRU,
+	}
+
+	if (len(nodeOptions.GPUVendors) > 0 || len(nodeOptions.GPUDevices) > 0) && nodeOptions.HasGPUs == 0 {
+		// at least one gpu in case the user didn't provide the amount
+		nodeOptions.HasGPUs = 1
```
xmonader: The variable name is misleading, should have been NumGPU

farmerbot/internal/find_node.go
```diff
@@ -0,0 +1,154 @@
...
+
+		possibleNodes = append(possibleNodes, node)
+	}
+
+	if len(possibleNodes) == 0 {
+		return 0, fmt.Errorf("could not find a suitable node with the given options: %+v", nodeOptions)
+	}
+
+	// Sort the nodes on power state (the ones that are ON first then waking up, off, shutting down)
+	sort.Slice(possibleNodes, func(i, j int) bool {
```
xmonader, line 97: Why do we need to sort by ID first?

farmerbot/internal/find_node.go
```diff
@@ -0,0 +1,154 @@
...
+		return possibleNodes[i].ID < possibleNodes[j].ID
+	})
+
+	sort.Slice(possibleNodes, func(i, j int) bool {
+		return possibleNodes[i].powerState < possibleNodes[j].powerState
+	})
```
xmonader: you can do the sorting in one call 

return ith powerstate < jth powerstate || ith.ID < jth.ID I think

farmerbot/internal/find_node.go
```diff
@@ -0,0 +1,154 @@
...
+	// update claimed resources
+	err := f.updateNode(nodeFound)
+	if err != nil {
+		return 0, fmt.Errorf("failed to power on found node %d", nodeFound.ID)
+	}
+
+	return uint32(nodeFound.ID), nil
+}
+
+// FilterIncludesSubStr filters a string slice according to if elements include a sub string
```
xmonader: leftover comment from another function? should be filterGPUs

farmerbot/internal/find_node.go
```diff
@@ -0,0 +1,154 @@
...
+func filterGPUs(gpus []zos.GPU, vendorsOrDevices []string, device bool) (filtered []zos.GPU) {
```
xmonader: this should have been splitted to two functions imo filterGPUsByVendors and filterGPUsByDevices

farmerbot/internal/farmerbot.go
```diff
@@ -0,0 +1,425 @@
...
+			return nil, fmt.Errorf("failed to load request payload: %w", err)
+		}
+
+		if err := f.powerOn(subConn, nodeID); err != nil {
+			return nil, fmt.Errorf("failed to power on node %d: %w", nodeID, err)
+		}
+
+		// Exclude node from farmerbot management
+		// (It is not allowed if we tried to power off a node the farmer decided to power on)
+		// the farmer should include it again if he wants to the bot to manage it
```
xmonader, line 212: So at this point this node is excluded even if the farmer defined it in their config?

farmerbot/internal/farmerbot.go
```diff
@@ -0,0 +1,425 @@
...
+	}
+	defer subConn.Close()
+
+	log.Debug().Msg("Fetch nodes")
+	farmNodes, err := subConn.GetNodes(uint32(f.state.farm.ID))
+	if err != nil {
+		return err
+	}
+
+	// remove nodes that doesn't exist anymore in the farm
```
xmonader: don't exist on the farm or the farmer didn't include them in the range specified?

farmerbot/internal/farmerbot.go
```diff
@@ -0,0 +1,425 @@
...
+	log.Debug().Msg("power management")
+	err = f.manageNodesPower(subConn)
+	if err != nil {
+		log.Error().Err(err).Msg("failed to power management nodes")
+	}
+
+	return nil
+}
+
+func (f *FarmerBot) addOrUpdateNode(ctx context.Context, subConn Sub, nodeID uint32) (node, error) {
```
xmonader: I'm confused with this one, Why not update it directly?

farmerbot/internal/farmerbot.go
```diff
@@ -0,0 +1,425 @@
...
+		// we also do not go through the code if we have woken up too many nodes at once => subtract (10 * (n-1))/min(periodic_wake up_limit, amount_of_nodes) from 8460
+		// now we can divide that by 10 and randomly generate a number in that range, if it's 0 we do the random wake up
+		log.Info().Uint32("nodeID", uint32(node.ID)).Msg("Random wake up")
+		return true
+	}
+
+	return false
+}
+
+func (f *FarmerBot) validateAccountEnoughBalance(sub *substrate.Substrate) error {
```
xmonader: This definitely needs to be called during the farmerbot start, we shouldn't let them start it if they don't have enough balance to maintain it running for a month at least

xmonader (changes requested):
Review is finished, I still have some questions on the behavior, but all in all: great job 💯. We are still missing the part of running this as a docker container too.

farmerbot/README.md
rawdaGastan: It's `ID` because it is an item of the list

farmerbot/internal/types.go
rawdaGastan: The order doesn't reflect anything

farmerbot/internal/types.go
rawdaGastan: Node resources don't include `total`, `used`, `system` resources

farmerbot/internal/types.go
rawdaGastan: I cannot rename as `MarshalText` is overridden when I marshal the YAML content

farmerbot/internal/constants.go
rawdaGastan: Yes, what is the problem here? 👀

farmerbot/cmd/root.go
rawdaGastan: It already has a default

farmerbot/cmd/root.go
rawdaGastan: No file restrictions

farmerbot/internal/state.go
rawdaGastan: yes

farmerbot/internal/farmerbot.go
rawdaGastan: No, If it was included and for any reason, the node doesn't exist anymore in the farm.

farmerbot/internal/farmerbot.go
rawdaGastan: It is mentioned in the README

farmerbot/internal/farmerbot.go
rawdaGastan: As per our discussion, we will set 100 TFT as a minimum balance allowed for running the bot.

farmerbot/internal/find_node.go
```diff
@@ -0,0 +1,157 @@
...
+	})
+
+	nodeFound := possibleNodes[0]
+	log.Debug().Uint32("nodeID", uint32(nodeFound.ID)).Msg("Found a node")
+
+	// claim the resources until next update of the data
+	// add a timeout (after 30 minutes we update the resources)
+	nodeFound.timeoutClaimedResources = time.Now().Add(timeoutPowerStateChange)
+
+	if nodeOptions.Dedicated || nodeOptions.NumGPU > 0 {
```
mariobassem, line 109: why do we reserve a node's total resources if there are requested gpus?

farmerbot/internal/find_node.go
```diff
@@ -0,0 +1,157 @@
...
+		return 0, fmt.Errorf("could not find a suitable node with the given options: %+v", nodeOptions)
+	}
+
+	// Sort the nodes on power state (the ones that are ON first then waking up, off, shutting down)
+	sort.Slice(possibleNodes, func(i, j int) bool {
+		return possibleNodes[i].powerState < possibleNodes[j].powerState ||
+			possibleNodes[i].ID < possibleNodes[j].ID // for tests
+	})
+
+	nodeFound := possibleNodes[0]
```
mariobassem, line 102: i think we still need the possible nodes list. a power on call for the first node may fail for any reason, and maybe it's best to check with other suitable nodes if that's the case

farmerbot/internal/find_node.go
```diff
@@ -0,0 +1,157 @@
...
+			if node.dedicated && nodeOptionsCapacity != node.resources.total {
+				continue
+			}
+		}
+
+		if slices.Contains(nodeOptions.NodesExcluded, uint32(node.ID)) {
+			continue
+		}
+
+		if !node.canClaimResources(nodeOptionsCapacity, f.config.Power.OverProvisionCPU) {
```
mariobassem, line 85: how do we make sure that the local state of nodes is not invalidated somehow? a user could deploy contracts on any of the monitored nodes and consume it's resources, which invalidates the state, no?

farmerbot/internal/farmerbot.go
```diff
@@ -0,0 +1,434 @@
...
+		}
+
+		log.Debug().Uint32("nodeID", nodeID).Msg("Node is updated with latest changes successfully")
+		return nodeObj, nil
+	}
+
+	// if node doesn't exist, we should add it
+	nodeObj, err := getNode(ctx, subConn, f.rmbNodeClient, nodeID, neverShutDown, false, f.state.farm.DedicatedFarm, on)
+	if err != nil {
+		if nodeObj.powerState == on {
```
mariobassem: any error from getNode will return a zero value for node, which in turn will always have `powerState == on` since `on` -> 0

farmerbot/internal/farmerbot.go
```diff
@@ -0,0 +1,434 @@
...
+			return node{}, fmt.Errorf("node %d is not responding while we expect it to: %w", nodeID, err)
+		}
+		log.Error().Err(err).Uint32("nodeID", nodeID).Msg("Failed to get node")
```
mariobassem: the caller of addOrUpdateNode logs this error, no?

farmerbot/internal/farmerbot.go
```diff
@@ -0,0 +1,434 @@
...
+	farmerTwinID, err := subConn.GetTwinByPubKey(f.identity.PublicKey())
+	if err != nil {
+		return err
+	}
+
+	farmRouter.WithHandler("version", func(ctx context.Context, payload []byte) (interface{}, error) {
+		return version.Version, nil
+	})
+
+	nodeRouter.WithHandler("findnode", func(ctx context.Context, payload []byte) (interface{}, error) {
```
mariobassem, line 123: why not add some rate limiting for twins that call findnode? since includenode, poweroff, and poweron are only accessed by the farm owner
also, maybe we could somehow add input limitations in the limiter: a large list of gpu vendors/devices will make the farmerbot busy looking for a match

mariobassem (commented):
a really great work.
i'm yet to finish the reivew, but i have some small questions and notes for now

farmerbot/internal/farmerbot.go
xmonader: That's a good improvement i'll open a separate issue for that one

rawdaGastan: I think we should open an issue for it

farmerbot/internal/find_node.go
xmonader: Problem here is, booting another node is also subject to the same thing and can get worse for the rest of the offline nodes even. I don't think that should be the behavior, the node may take 1-30 mins too boot, while the farmerbot iteration is every 5mins :D

farmerbot/internal/find_node.go
xmonader: the state is rebuilt every 5mins, we can have a 10% buffer around it for optimistic calculations

xmonader: will always get real time

farmerbot/internal/find_node.go
rawdaGastan: Because requesting GPUs should request total capacity (node will be rented)

farmerbot/internal/find_node.go
rawdaGastan: https://github.com/threefoldtech/tfgrid-sdk-go/issues/571

xmonader (approved):

Issue #322, opened 2023-08-13
improve the gpu indexing with the change in number of nodes


xmonader:
Right now we run the indexing goroutine every 1hr (it's configurable), however maybe we should also do better if the number of the nodes changes before and after? and we can even improve by indexing  the newest ones first?

Omarabdul3ziz:
we kept the periodically running ticker to requery the nodes and upsert the GPUs info each hour.
also for newly added nodes a Postgres trigger is added on the node table which will notify the client with the added node twin id and then the client will ask the node with an RMB call if it has a GPU or not.

but this focus on enhancing the indexing process for newly added nodes, not for the nodes updated with add/remove gpu,  also considered using a trigger depending on another field that changes whenever a new GPU is added, even if it's added to a node already in the table. but couldn't find something like that

do you have any suggestions or immediately indexing the new node is enough?

A-Harby:
When a new node is added with a GPU, the Postgres database is triggered with notify_nodes_count_changed, and the node data is added.

For enhancement, it will be more visible on the mainnet, and if more improvement is needed, reopen the issue.

Omarabdul3ziz:
Update:
had to reopen this because the implementation wasn't stable, had a bit more complexity than needed, and relied on a postgres feature.
we rolled back the code using these two prs #475 and #480.
and a new release `v0.11.6` without these changes is now deployed.

Note:
consider checking the last row in the node table and, if it's changed, trigger a reindex

ramezsaeed:
Current implementation:
- each 1 hour the indexer will call all the nodes to get its' gpu info if it has one
- each 5 min a check will run to watch the node table count if it is changed it will trigger the indexer to run, if not it will do nothing

ramezsaeed:
Test in Progress:
Add Test coverage here:
https://app.testlodge.com/a/26076/projects/40893/suites/228929?expand_section=423253#suite_section_423253

PR #512, opened 2023-11-26
Check for new added nodes

AbdelrahmanElawady:
### Description

Check for new added nodes periodically and update GPU data when new nodes are added.

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/322

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/grid-proxy/internal/explorer/db/postgres.go
+++ b/grid-proxy/internal/explorer/db/postgres.go
@@ -935,8 +935,14 @@ func (p *PostgresDatabase) UpsertNodesGPU(ctx context.Context, nodesGPU []types.
 	return nil
 }
 
-func (p *PostgresDatabase) GetLastNodeTwinID(ctx context.Context) (uint32, error) {
+func (p *PostgresDatabase) GetLastNodeTwinID(ctx context.Context) (int64, error) {
 	var node Node
 	err := p.gormDB.Table("node").Order("twin_id DESC").Limit(1).Scan(&node).Error
-	return uint32(node.TwinID), err
+	return node.TwinID, err
+}
+
+func (p *PostgresDatabase) GetNodeTwinIDsAfter(ctx context.Context, twinID int64) ([]int64, error) {
+	nodeTwinIDs := make([]int64, 0)
+	err := p.gormDB.Table("node").Select("twin_id").Where("twin_id > ?", twinID).Order("twin_id DESC").Scan(&nodeTwinIDs).Error
+	return nodeTwinIDs, err
 }
--- a/grid-proxy/internal/explorer/db/postgres.go
+++ b/grid-proxy/internal/explorer/db/postgres.go
@@ -934,3 +934,9 @@ func (p *PostgresDatabase) UpsertNodesGPU(ctx context.Context, nodesGPU []types.
 	}
 	return nil
 }
+
+func (p *PostgresDatabase) GetLastNodeTwinID(ctx context.Context) (uint32, error) {
+	var node Node
+	err := p.gormDB.Table("node").Order("twin_id DESC").Limit(1).Scan(&node).Error
+	return uint32(node.TwinID), err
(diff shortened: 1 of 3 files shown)


grid-proxy/internal/gpuindexer/gpuindexer.go
```diff
@@ -29,6 +30,7 @@ type NodeGPUIndexer struct {
 	batchSize              int
 	nodesGPUResultsChan    chan []types.NodeGPU
 	nodesGPUBatchesChan    chan []types.NodeGPU
+	newNodeTwinIDChan      chan []uint32
```
xmonader, line 33: Why is that a channel of int32 slice? this should keep receiving the latest twinID inserted on the nodes table,  no?

xmonader: ah forget it checks the nodes between the last id you checked and the last id on the table, could be more than one indeed

grid-proxy/internal/explorer/db/postgres.go
```diff
@@ -934,3 +934,9 @@ func (p *PostgresDatabase) UpsertNodesGPU(ctx context.Context, nodesGPU []types.
 	}
 	return nil
 }
+
+func (p *PostgresDatabase) GetLastNodeTwinID(ctx context.Context) (uint32, error) {
```
xmonader: Maybe we should change it to be GetGreaterThanTwinID and use the twinID we checked last? this should get us the list of ids that we can directly send?

AbdelrahmanElawady: Initially, we thought we only need the latest ID and all new nodes will be from last checked to the latest ID however this is not right as twin IDs are shared with other entities so node twin IDs don't have to be incremented one by one. So, this approach is probably better.

grid-proxy/internal/gpuindexer/gpuindexer.go
```diff
@@ -183,10 +195,42 @@ func (n *NodeGPUIndexer) Start(ctx context.Context) {
...
+
 	log.Info().Msg("GPU indexer started")
 
 }
 
+func (n *NodeGPUIndexer) watchNodeTable(ctx context.Context) {
+	ticker := time.NewTicker(newNodesCheckInterval)
+	latestCheckedID, err := n.db.GetLastNodeTwinID(ctx)
+	if err != nil {
+		log.Error().Err(err).Msg("failed to get last node twin id")
```
xmonader, line 208: This means the proxy is having a seriou issue where there're no nodes in the table? or could that while the pod is starting for the first time?Is that the reason why we don't panic here?

grid-proxy/internal/gpuindexer/gpuindexer.go
```diff
@@ -183,10 +195,42 @@ func (n *NodeGPUIndexer) Start(ctx context.Context) {
...
+			}
+			nodeTwinIDs := make([]uint32, 0)
+			for _, id := range newIDs {
+				nodeTwinIDs = append(nodeTwinIDs, uint32(id))
+			}
+
+			n.newNodeTwinIDChan <- nodeTwinIDs
+			latestCheckedID = int64(nodeTwinIDs[0])
+		case <-ctx.Done():
+			break
```
xmonader: should it be return?

grid-proxy/internal/gpuindexer/gpuindexer.go
AbdelrahmanElawady: I think there might be non-fatal errors here that would not stop the operation of the indexer, that's why we don't panic. What I can think of is maybe some network issue or db connection error on start. I think we don't need to panic here.

xmonader (approved):

xmonader:
Thank you 💯

PR #520, opened 2023-11-29
adds a region filter on nodes/farms endpoints

### LINKED ISSUE #508: Add filter to support filtering farms/nodes with region

Omarabdul3ziz:
- updates the generate script to add some countries objects.
- updates the filter types.
- joins the country table to node/farm table, and updates sql query.
- update the mock_client wiht the new filter.
- update swagger docs

- fix: #508

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/grid-proxy/tools/db/generate.go
+++ b/grid-proxy/tools/db/generate.go
@@ -49,7 +49,19 @@ const (
 
 var (
 	countries = []string{"Belgium", "United States", "Egypt", "United Kingdom"}
-	cities    = map[string][]string{
+	regions   = map[string]string{
+		"Belgium":        "Europe",
+		"United States":  "Americas",
+		"Egypt":          "Africa",
+		"United Kingdom": "Europe",
+	}
+	countriesCodes = map[string]string{
+		"Belgium":        "BG",
+		"United States":  "US",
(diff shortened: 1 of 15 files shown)


grid-proxy/tools/db/generate.go
```diff
@@ -49,7 +49,19 @@ const (
 
 var (
 	countries = []string{"Belgium", "United States", "Egypt", "United Kingdom"}
-	cities    = map[string][]string{
+	regions   = map[string]string{
```
muhamadazmy, line 52: I don't get this, are these the only countries, and cities that we know about ?

muhamadazmy: Shouldn't we use a real countries db instead?

Omarabdul3ziz: The `generate.go` script generates a dump of data for the tests to work on, so it just shuffles between these four countries and selects the region/city/code for this country name to fill the fields on the node/farm table. 
then, the tests work on filtering the resources (nodes/farms) based on these loaded dump data.

muhamadazmy (approved):

Issue #507, opened 2023-11-22
add encoding/decoding formats as an option to the peer


rawdaGastan:
- Add peer encoding/decoding format and schema which means the user has to specify this as a value during creating a new peer.
- Also make sure to move the NewPeer functions args to a new struct `PeerOptions` to deal with arguments and default values.

suggestion: https://golang.cafe/blog/golang-functional-options-pattern.html

AbdelrahmanElawady:
Update:

- Peer now takes non-required arguments through options.
- Peer now uses an Encoder interface to manage encoding payloads.
- The interface allows for more schemas support in the future.

PR #526, opened 2023-12-04
Add encoding options to peer

AbdelrahmanElawady:
### Description

Add encoding options to RMB Peer and defaults for Peer arguments.

### Related Issues

- #507 

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation


rmb-sdk-go/peer/encoder/json.go
```diff
@@ -0,0 +1,21 @@
...
+
+func (e *jsonEncoder) Encode(v interface{}) ([]byte, error) {
+	return json.Marshal(v)
+}
+
+func (e *jsonEncoder) Decode(data []byte, out interface{}) error {
+	return json.Unmarshal(data, out)
+}
+
+func newJSONEncoder() Encoder {
```
muhamadazmy: This should be public IMHO, even if it's the default the user should still explictely pass it as an option

rmb-sdk-go/peer/peer.go
```diff
@@ -80,14 +81,28 @@ func getIdentity(keytype string, mnemonics string) (substrate.Identity, error) {
...
-	enableEncryption bool,
-	handler Handler) (*Peer, error) {
-	identity, err := getIdentity(keytype, mnemonics)
+	handler Handler,
+	opts ...PeerOpt) (*Peer, error) {
+
+	cfg := &peerCfg{
+		relayURL:         "wss://relay.grid.tf",
+		session:          "",
+		enableEncryption: false,
```
muhamadazmy: encryption should be `true` by default actually. only disable explicetly via the option

rmb-sdk-go/peer/peer.go
```diff
@@ -80,14 +81,28 @@ func getIdentity(keytype string, mnemonics string) (substrate.Identity, error) {
...
+		schema:           encoder.DefaultSchema,
+	}
+	for _, o := range opts {
+		o(cfg)
+	}
+
+	e, err := encoder.NewEncoder(cfg.schema)
+	if err != nil {
+		return nil, err
+	}
```
muhamadazmy: This is not the right way. This basically means that NewEncoder need to be modified everytime you have to implement a new encoding type. 

What if in another app i wanna use this with "binary" encoding, do i have to fork and modify your library before i can do this ?

The correct way to do this instead is that your option instead must be  `WithEncoder` (or something) that takes an instance of Encoder type. 

If encoder is not set (yourc `cfg.encoder` is still nil after applying all the Withs) then you can then create the default JsonEncoder and use that.

rmb-sdk-go/peer/peer.go
```diff
@@ -147,13 +162,58 @@ func NewPeer(
...
+	return func(p *peerCfg) {
+		// to ensure only ed25519 and sr25519 are used
+		if keyType != KeyTypeEd25519 {
+			keyType = KeyTypeSr25519
+		}
+		p.keyType = keyType
+	}
+}
+
+func WithEncoder(schema encoder.Schema) PeerOpt {
```
muhamadazmy: Check the other comment, this instead should take `Encoder` not Schema! to make it easier for developer to implement their own encoder types. without the need to modify this library code

rmb-sdk-go/peer/peer.go
```diff
@@ -147,13 +162,58 @@ func NewPeer(
 		reader:  reader,
 		writer:  writer,
 		handler: handler,
+		encoder: e,
```
muhamadazmy: I think the Peer need to also may have function `Encoder() Encoder` that returns the encoder. This allows handler to then use the encode to `decode` payloads. if they match expected schema.

```go
func (p *Peer) Encoder() Encoder {
   return p.encoder
}
```

rmb-sdk-go/peer/encoder/encoder.go
```diff
@@ -0,0 +1,30 @@
...
+type Schema string
+
+// supported schemas
+const (
+	JSONSchema    = "application/json"
+	DefaultSchema = JSONSchema
+)
+
+// NewEncoder returns an encoder based on the given schema
+func NewEncoder(schema Schema) (Encoder, error) {
```
muhamadazmy: Check the other comment regarding how the encoder need to be configured. 

In short, This means that user of this library cannot use custom encoding unless this library is modified (or forked) Instead the `WithEncoder` need to take an encoder object, not schema. Hence I (as a 3rd party) can implement my cutsom encoder type.

muhamadazmy (changes requested):

xmonader (approved):

muhamadazmy (approved):

PR #527, opened 2023-12-05
create mass deployment

Eslam-Nawara:
## Description:
- Create mass deployer package that deploys batch of vms according to a given configurations file
### Changes

- Add tfrobot package
- Add tfrobot to lint, test and build workflows and goreleaser
- Fix deployer skipping error problem in batch deploy in deployer
- Fix deployer return error on only `substrate.ErrAccountNotFound` in deployer validation

## Related Issues:
- #604 

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation


grid-proxy/pkg/client/grid_client.go
```diff
@@ -425,13 +425,13 @@ func (g *Clientimpl) ContractBills(ctx context.Context, contractID uint32, limit
...
 	return &http.Client{
-		Timeout: time.Second * 30,
+		Timeout: time.Second * 100,
 		Transport: &http.Transport{
 			DialContext: (&net.Dialer{
 				Timeout: time.Second,
 			}).DialContext,
 			TLSHandshakeTimeout:   time.Second,
-			ResponseHeaderTimeout: 5 * time.Second,
+			ResponseHeaderTimeout: 100 * time.Second,
```
Eslam-Nawara: This is required until 3.13 is on mainnet as the response takes too long rn, but fixed on devnet

grid-client/examples/mass_deployment/helpers.go
```diff
@@ -0,0 +1,108 @@
...
+			lock.Lock()
+			nodesIDs = append(nodesIDs, uint32(node.NodeID))
+			lock.Unlock()
+		}(node)
+	}
+	wg.Wait()
+	return nodesIDs
+}
+
+func getNodesWithValidFileSystem(nodes []uint32, tfPluginClient deployer.TFPluginClient, ctx context.Context) []uint32 {
```
Eslam-Nawara: This is required until 3.13 is on mainnet, fixed on devnet with https://github.com/threefoldtech/zos/pull/2143

Eslam-Nawara:
Tested on batch of 100 vms and the deployment time is very good for now
image

.github/workflows/build.yml
```diff
@@ -28,6 +28,7 @@ jobs:
           - monitoring-bot
           # - rmb-sdk-go
           - user-contracts-mon 
+          # - mass-deployer
```
rawdaGastan: add its build and docker image 👍🏻

grid-client/examples/mass_deployment/helpers.go
```diff
@@ -0,0 +1,140 @@
...
+	"sync"
+
+	"github.com/threefoldtech/tfgrid-sdk-go/grid-client/deployer"
+	"github.com/threefoldtech/tfgrid-sdk-go/grid-proxy/pkg/types"
+)
+
+var (
+	statusUp  = "up"
+	minRootfs = *convertGBToBytes(2)
+	twn       = uint64(5432)
```
rawdaGastan: what is the twn?

grid-client/examples/mass_deployment/helpers.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if len(nodesIDs) < totalVMCount {
+		return []uint32{}, fmt.Errorf("no available nodes found, Only found %d\n", len(nodesIDs))
+	}
+
+	fmt.Printf("Found free %d nodes!\n", len(nodes))
+	return nodesIDs[:totalVMCount], nil
+}
+
+func getReachableNodes(nodes []uint32, tfPluginClient deployer.TFPluginClient, ctx context.Context) []uint32 {
```
rawdaGastan: is it required?

mass-deployer/Makefile
```diff
@@ -0,0 +1,65 @@
...
+
+getverifiers:
+	@echo "Installing staticcheck" && go get -u honnef.co/go/tools/cmd/staticcheck && go install honnef.co/go/tools/cmd/staticcheck
+	@echo "Installing gocyclo" && go get -u github.com/fzipp/gocyclo/cmd/gocyclo && go install github.com/fzipp/gocyclo/cmd/gocyclo
+	@echo "Installing deadcode" && go get -u github.com/remyoudompheng/go-misc/deadcode && go install github.com/remyoudompheng/go-misc/deadcode
+	@echo "Installing misspell" && go get -u github.com/client9/misspell/cmd/misspell && go install github.com/client9/misspell/cmd/misspell
+	@echo "Installing golangci-lint" && go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.2
+	go mod tidy
+
+verifiers: fmt lint cyclo deadcode spelling staticcheck
```
rawdaGastan: now we only use `golangci-lint`

mass-deployer/README.md
```diff
@@ -0,0 +1,59 @@
+# mass-deployer
+
+Mass Deplyer tool designed to automate mass deployment of batchs of VMs on ThreeFold Grid.
```
rawdaGastan: batchs -> groups (to match requirements)

mass-deployer/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Features
+
+-   **Mass Deployment:** Deploy groups of vms on ThreeFold Grid simultaneously.
+-   **Customizable Configurations:** Define Node groups, VMs groups and other configurations through YAML or JSON file.
```
rawdaGastan: let it be YAML only

mass-deployer/README.md
```diff
@@ -0,0 +1,59 @@
...
+  - name: example-vm
+    vms_count: 2
+    node_group: example-group
+    cpu: 2
+    mem: 4096
+    flist: example-flist,
+    entry_point: /sbin/zinit init
+    # ... other fields
+
+sshkey: example-ssh-key
```
rawdaGastan: We have agreed to have the ssh key with its name (to allow more than one ssh key)

mass-deployer/README.md
```diff
@@ -0,0 +1,59 @@
...
+> Make sure to replace placeholders and adapt the groups based on your actual project details.
+
+3.  Run the deployer with path to the config file
+```bash
+$ mass-deployer -c path/to/your/config.yaml
+```
+
+## Download
+
+-   Download the binaries from [releases](https://github.com/threefoldtech/tfgrid-sdk-go/releases)
```
rawdaGastan: add in goreleaser file

mass-deployer/cmd/main.go
```diff
@@ -0,0 +1,91 @@
...
+	flag.StringVar(&configFile, "c", "", "path to config file")
+	flag.Parse()
+
+	if configFile == "" {
+		log.Fatal("couldn't locate config file")
+	}
+
+	cfg, err := parser.ParseConfig(configFile)
+	if err != nil {
+		log.Fatal(err)
```
rawdaGastan: wrap the errors
check all

mass-deployer/cmd/main.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Minute)
+	defer cancel()
+
+	groupsNodes := map[string][]int{}
+
+	pass := map[string]error{}
```
rawdaGastan: I don't understand the pass map

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,45 @@
...
+	"errors"
+	"os"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+func ParseConfig(filtName string) (Config, error) {
+	str := strings.Split(filtName, ".")
+	if len(str) < 2 {
```
rawdaGastan: you can use `filepath.Ext()`

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,45 @@
...
+		return Config{}, errors.New("invalid config file extension")
+	}
+
+	conf := Config{}
+	if str[len(str)-1] == "json" {
```
rawdaGastan: remove json :D

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,45 @@
...
+		configFile, err := os.ReadFile(filtName)
+		if err != nil {
+			return Config{}, err
+		}
+
+		err = yaml.Unmarshal(configFile, &conf)
+		if err != nil {
+			return Config{}, err
+		}
+		return Config{}, nil
```
rawdaGastan: why returning `Config{}`?

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	return Config{}, errors.New("invalid config file extension")
```
rawdaGastan: We can return the YAML configs instead

mass-deployer/README.md
```diff
@@ -0,0 +1,59 @@
+# mass-deployer
```
rawdaGastan: add coverage badge

mass-deployer/internal/parser/structs.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+type Vm struct {
+	Name        string `yaml:"name" json:"name"`
+	Count       int    `yaml:"vms_count" json:"vms_count"`
+	Nodegroup   string `yaml:"node_group" json:"node_group"`
+	FreeCPU     int    `yaml:"cpu" json:"cpu"`
+	FreeMRU     int    `yaml:"mem" json:"mem"`
+	Disk        Disk   `yaml:"ssd" json:"ssd"`
+	HDDAttached bool   `yaml:"hdd" json:"hdd"`
```
rawdaGastan: Couldn't understand it ? 0_0

mass-deployer/internal/parser/structs.go
```diff
@@ -0,0 +1,44 @@
+package parser
+
+type Config struct {
+	NodeGroups []NodesGroup `yaml:"node_groups" json:"node_groups"`
+	Vms        []Vm         `yaml:"vms" json:"vms"`
+	SSHKey     string       `yaml:"sshkey" json:"sshkey"`
```
rawdaGastan: make it a struct with name and key

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/threefoldtech/tfgrid-sdk-go/mass-deployer/internal/parser"
+	"github.com/threefoldtech/zos/pkg/gridtypes"
+)
+
+type Deployer struct {
+	TFPluginClient deployer.TFPluginClient
+}
+
+func NewDeployer(conf parser.Config) (Deployer, error) {
+	network := os.Getenv("NETWORK")
```
rawdaGastan: why getting from env? aren't mnemonic and network in our configs

rawdaGastan (changes requested):
Good job 🔥

mass-deployer/README.md
Eslam-Nawara: so it's going to be a list of sshkeys?

mass-deployer/cmd/main.go
Eslam-Nawara: I've changed it to a list of groups names for now

mass-deployer/internal/parser/parser.go
Eslam-Nawara: not sure what you mean here

.goreleaser.yaml
```diff
@@ -32,6 +32,22 @@ builds:
...
+      - linux
+      - windows
+      - darwin
+    binary: mass-deployer
+    id: mass-deployer
+
+    ignore:
+      - goos: windows
+    ldflags:
+      - -X github.com/threefoldtech/tfgrid-sdk-go/grid-cli/cmd.version={{.Tag}}
```
rawdaGastan: The tag and the commit should b related to version and commit variables for the mass deployer

this means that the version variable in cmd package of grid-cli will have a value of the tag

mass-deployer/Makefile
```diff
@@ -0,0 +1,36 @@
...
+	go mod tidy
+
+lint:
+	@echo "Running $@"
+	golangci-lint run -c ../.golangci.yml
+
+build:
+	@echo "Running $@"
+	@go build -ldflags=\
+	"-X 'github.com/threefoldtech/tfgrid-sdk-go/mass-deployer/cmd.commit=$(shell git rev-parse HEAD)'\
```
rawdaGastan, line 20: you did it here, please change in the goreleaser

mass-deployer/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+<a href='https://github.com/jpoles1/gopherbadger' target='_blank'>![gopherbadger-tag-do-not-edit](https://img.shields.io/badge/Go%20Coverage-90%25-brightgreen.svg?longCache=true&style=flat)</a>
+
+Mass Deplyer tool designed to automate mass deployment of groups of VMs on ThreeFold Grid.
```
rawdaGastan: `is designed`

mass-deployer/README.md
```diff
@@ -0,0 +1,66 @@
...
+$ mass-deployer -c path/to/your/config.yaml
+```
+
+## Using Docker
+```bash
+docker build -t mass-deployer -f Dockerfile ../
+docker run -v $(pwd)/config.yaml:/config.yaml -it mass-deployer:latest -c /config.yaml
+```
+
+## Build
```
rawdaGastan, line 62: add test also

mass-deployer/README.md
```diff
@@ -0,0 +1,66 @@
...
+    entry_point: /sbin/zinit init
+    # ... other fields
+
+sshkey: 
+  example1: ssh-key1
+mnemonic: example-mnemonic
+network: example-network
+```
+
+> You can use this [example](./docs/conf.yaml) for further guidance, and make sure to replace placeholders and adapt the groups based on your actual project details.
```
rawdaGastan: Cannot find docs directory, also you can use the example directly instead of the yaml part

mass-deployer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+	Short: "A tool for deploying groups of vms on Threefold Grid",
+
+	Run: func(cmd *cobra.Command, args []string) {
+		configPath, err := cmd.Flags().GetString("config")
+		if err != nil || configPath == "" {
+			log.Fatal().Err(err).Msg("error in config file")
+			return
+		}
+
+		if filepath.Ext(configPath) != ".yaml" {
```
rawdaGastan: It can be `yml`

mass-deployer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+			log.Fatal().Err(err).Msg("invalid config file extension")
```
rawdaGastan: Add the wrong extension to the error

mass-deployer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		configFile, err := os.ReadFile(configPath)
+		if err != nil {
+			log.Fatal().Err(err).Msg("failed to open config file")
+		}
+
+		err = runDeployer(configFile)
+		if err != nil {
+			log.Fatal().Err(err).Msg("failed to parse config file")
```
rawdaGastan: wrong error msg

mass-deployer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/threefoldtech/tfgrid-sdk-go/grid-client/workloads"
+	"github.com/threefoldtech/tfgrid-sdk-go/mass-deployer/internal/parser"
+	deployer "github.com/threefoldtech/tfgrid-sdk-go/mass-deployer/pkg/mass-deployer"
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "mass-deployer",
+	Short: "A tool for deploying groups of vms on Threefold Grid",
+
+	Run: func(cmd *cobra.Command, args []string) {
```
rawdaGastan, line 19: you can use RunE and return the errors

mass-deployer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+		if err != nil {
+			log.Fatal().Err(err).Msg("failed to parse config file")
+		}
+	},
+}
+
+func Execute() {
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
```
rawdaGastan, line 47: I f you changed to RunE you can fatal the returned error

mass-deployer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+
+	d, err := deployer.NewDeployer(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to create deployer: %v", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Minute)
+	defer cancel()
+
+	groupsNodes := map[string][]int{}
```
rawdaGastan: shouldn't we move the logic to deployer pkg?

mass-deployer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to parse config file: %v", err)
+	}
+
+	d, err := deployer.NewDeployer(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to create deployer: %v", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Minute)
```
rawdaGastan: why setting a context with timeout?

mass-deployer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+	for group, info := range pass {
+
+		groupInfo, err := yaml.Marshal(info)
+		if err != nil {
+			log.Debug().Err(err).Msg("failed to marshal json")
+		}
+		fmt.Printf("%s: \n%v\n", group, string(groupInfo))
+	}
+
+	fmt.Println("error:")
```
rawdaGastan: don't print error if we didn't get errors

mass-deployer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+				fail[group] = err
+			} else {
+				pass[group] = info
+			}
+		}(group, vms)
+	}
+	wg.Wait()
+
+	fmt.Println("deployment took ", time.Since(deploymentStart))
+	fmt.Println("ok:")
```
rawdaGastan: don't print ok if pass is empty

mass-deployer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+	for group, info := range pass {
```
rawdaGastan: `groupName` would be better

mass-deployer/example/conf.yaml
```diff
@@ -0,0 +1,33 @@
+node_groups:
+  - name: group_a
+    nodes_count: 3 # amount of nodes to be found
+    free_cpu: 2 # number of logical cores
+    free_mru: 16384 # amount of memory in mb
```
rawdaGastan: make `mb` -> `MB` and apply for others

mass-deployer/example/conf.yaml
```diff
@@ -0,0 +1,33 @@
...
+    free_ssd: 100 # amount of ssd storage in gb
+    free_hdd: 50 # amount of hdd storage in gb
+    dedicated: false # is dedicated nodes?
```
rawdaGastan: `are nodes dedicated`

mass-deployer/example/conf.yaml
```diff
@@ -0,0 +1,33 @@
...
+    pubip4: false # should the nodes have free ip v4
+    pubip6: false # should the nodes have free ip v6
+    certified: false # should the nodes be certified(if false the nodes could be certified of diyed) 
+    region: "europe" # region could be the name of the continents the nodes are located in (eg: europe, asia,...)
```
rawdaGastan: provide all values

mass-deployer/example/conf.yaml
```diff
@@ -0,0 +1,33 @@
...
+    min_bandwidth_ms: 100
```
rawdaGastan: What is `min_bandwidth_ms`?

mass-deployer/example/conf.yaml
```diff
@@ -0,0 +1,33 @@
...
+vms:
+  - name: examplevm
+    vms_count: 5 # amount of vms with the same configurations
+    node_group: group_a # the name of the predefined group of nodes
+    cpu: 1 # number of logical cores
+    mem: 256 # amount of memory in mb
+    ssd: # list of ssd storage needed to be mounted to the vm
+      capacity: 5 # capacity in gb
```
rawdaGastan: why `capacity` not `size`?

mass-deployer/example/conf.yaml
```diff
@@ -0,0 +1,33 @@
...
+      mount_point: /mnt/ssd
+    pubip4: false
```
rawdaGastan: continue the descriptions :D

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,89 @@
...
+	if len(conf.NodeGroups) == 0 {
+		return Config{}, fmt.Errorf("couldn't find any nodes groups to use in deployment")
+	}
+	if len(conf.Vms) == 0 {
+		return Config{}, fmt.Errorf("couldn't find any vms that need to be deployed")
+	}
+	if len(conf.SSHKeys) == 0 {
+		return Config{}, fmt.Errorf("user ssh_keys is invalid, ssh_keys shouldn't be empty")
+	}
+	if len(conf.Mnemonic) == 0 {
```
rawdaGastan: Please validate the mnemonic, also when checking a length of a string we should trim spaces of it `apply for all`
you can use `bip39.IsMnemonicValid(mnemonic)

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,89 @@
...
+		return Config{}, fmt.Errorf("user mnemonics is invalid, mnemonics shouldn't be empty")
```
rawdaGastan: mnemonics -> mnemonic

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,89 @@
...
+	networks := []string{"dev", "test", "qa", "main"}
+	if !slices.Contains(networks, conf.Network) {
+		return Config{}, fmt.Errorf("netwok %s is invalid", conf.Network)
+	}
+
+	for _, nodeGroup := range conf.NodeGroups {
+		if nodeGroup.Name == "" {
+			return Config{}, fmt.Errorf("node groups name is invalid, shouldn't be empty")
+		}
+		if nodeGroup.NodesCount <= 0 {
```
rawdaGastan: add the wrong values in errors `apply for all`

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	for _, vm := range conf.Vms {
+		if vm.Name == "" {
+			return Config{}, fmt.Errorf("vms name is invalid, shouldn't be empty")
+		}
+		if vm.Count <= 0 {
+			return Config{}, fmt.Errorf("vms_count in vms: %s is invalid, should be a positive number", vm.Name)
+		}
+		if vm.Nodegroup == "" {
```
rawdaGastan: check if node group exists

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,89 @@
...
+		if nodeGroup.FreeMRU <= 0 {
+			return Config{}, fmt.Errorf("free_mru in node_group: %s is invalid, should be a positive number", nodeGroup.Name)
+		}
+	}
+
+	for _, vm := range conf.Vms {
+		if vm.Name == "" {
+			return Config{}, fmt.Errorf("vms name is invalid, shouldn't be empty")
+		}
+		if vm.Count <= 0 {
```
rawdaGastan: check if the count doesn't match the count of nodes group

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		if vm.FreeCPU <= 0 {
+			return Config{}, fmt.Errorf("cpu in vms: %s is invalid, should be a positive number", vm.Name)
+		}
+		if vm.FreeMRU <= 0 {
+			return Config{}, fmt.Errorf("mem in vms: %s is invalid, should be a positive number", vm.Name)
+		}
+		for _, disk := range vm.SSDDisks {
+			if disk.Capacity <= 0 {
+				return Config{}, fmt.Errorf("ssd disk capacity in vms: %s is invalid, should be a positive number", vm.Name)
```
rawdaGastan: in `vm`
apply for others

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+		}
+		if vm.Flist == "" {
+			return Config{}, fmt.Errorf("vms flist is invalid, shouldn't be empty")
+		}
+		if vm.Entrypoint == "" {
+			return Config{}, fmt.Errorf("vms entry_point is invalid, shouldn't be empty")
+		}
+		if _, ok := conf.SSHKeys[vm.SSHKey]; !ok {
+			return Config{}, fmt.Errorf("vms ssh_key is invalid, should be valid ssh_key refers to one of ssh_keys list")
```
rawdaGastan: should be valid name not ssh_key?

mass-deployer/internal/parser/parser_test.go
```diff
@@ -0,0 +1,154 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseConfig(t *testing.T) {
+	t.Run("invaild yaml format", func(t *testing.T) {
+		conf := ` {
+  "node_groups": [
```
rawdaGastan, line 53: could be a smaller example :D

mass-deployer/internal/parser/structs.go
```diff
@@ -0,0 +1,45 @@
+package parser
```
rawdaGastan: can we rename to types?
also why types is in parser pkg

mass-deployer/internal/parser/structs.go
```diff
@@ -0,0 +1,45 @@
...
+	Mnemonic   string            `yaml:"mnemonic"`
+	Network    string            `yaml:"network"`
+}
+
+type NodesGroup struct {
+	Name       string `yaml:"name"`
+	NodesCount uint64 `yaml:"nodes_count"`
+	FreeCPU    uint64 `yaml:"free_cpu"`
+	FreeMRU    uint64 `yaml:"free_mru"`
+	FreeSSD    uint64 `yaml:"free_ssd"`
```
rawdaGastan: FreeSRU, FreeHRU?

mass-deployer/pkg/mass-deployer/helpers.go
```diff
@@ -0,0 +1,17 @@
+package deployer
```
rawdaGastan: Don't name it with helpers, Thabet doesn't like it :D you can add it with deployer file or have a file for each o_o

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+type vmDeploymentInfo struct {
+	nodeID         uint32
+	vmName         string
+	deploymentName string
+}
+
+type VMInfo struct {
+	Name      string
```
rawdaGastan: Why this info exactly is used?

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,206 @@
...
+	PublicIP4 string
+	PublicIP6 string
+	YggIP     string
+	IP        string
+	Mounts    []workloads.Mount
+}
+
+func NewDeployer(conf parser.Config) (Deployer, error) {
+	network := conf.Network
+	log.Printf("network: %s", network)
```
rawdaGastan: use zerolog info for example

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,206 @@
...
+	if group.Dedicated {
+		filter.Dedicated = &group.Dedicated
+	}
+
+	freeSSD := []uint64{group.FreeSSD}
+	if group.FreeSSD == 0 {
+		freeSSD = nil
+	}
+	freeHDD := []uint64{group.FreeHDD}
+	if group.FreeSSD == 0 {
```
rawdaGastan: FreeHDD not FreeSSD

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,206 @@
...
+
+	for i, vm := range vms {
+		nodeID := nodes[i%nodesCount]
+		wg.Add(1)
+
+		go func(vm workloads.VM, i int, nodeID uint32) {
+			defer wg.Done()
+
+			network := workloads.ZNet{
+				Name:        generateRandomString(10),
```
rawdaGastan: we can use the vm name and add network to it

rawdaGastan (changes requested):

mass-deployer/example/conf.yaml
Eslam-Nawara: It's not implemented yet

mass-deployer/pkg/mass-deployer/helpers.go
xmonader: 😆 😆

mass-deployer/internal/parser/parser.go
Eslam-Nawara: The vms count can be not the same as nodes count I guess, I mean more than one vm can be deployed on the same node, right?

mass-deployer/internal/parser/parser.go
Eslam-Nawara: it's a group of vms so, vms sounds fine I think

mass-deployer/internal/parser/structs.go
Eslam-Nawara: they're the types the parser use, should they be in another place?

mass-deployer/pkg/mass-deployer/deployer.go
Eslam-Nawara: to export the passed deployments info to be displayed to the user

mass-deployer/pkg/mass-deployer/deployer.go
Eslam-Nawara: this is zerolog, not sure what you mean here

mass-deployer/pkg/mass-deployer/deployer.go
Eslam-Nawara: the vm name is used for multible vms so having network to be network+vm name would get conflict error

mass-deployer/example/conf.yaml
rawdaGastan: We can remove it as it is not supported and create an issue for it.

mass-deployer/internal/parser/parser.go
rawdaGastan: Let it be `vms group` because it confused me

mass-deployer/internal/parser/structs.go
rawdaGastan: Yes in deployer pkg. Parser shouldn't have anything more than just parsing

xmonader:
Let's make sure to support mnemonic and network as env variables too please

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,283 @@
...
+	err = tfPluginClient.DeploymentDeployer.BatchDeploy(ctx, deployemnts.vmDeployments)
+	if err != nil {
+		return []string{}, err
+	}
+	vmsInfo := loadDeploymentsInfo(tfPluginClient, deployemnts.deploymentsInfo)
+
+	return vmsInfo, nil
+}
+
+func buildDeployments(tfPluginClient deployer.TFPluginClient, vms []Vms, nodesIDs []int, sshKeys map[string]string) groupDeploymentsInfo {
```
AbdelrahmanElawady: `tfPluginClient` is not used here.

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,283 @@
...
+		vmsOfNodeGroups[vm.Nodegroup] = append(vmsOfNodeGroups[vm.Nodegroup], vm)
+	}
+
+	for nodeGroup, vms := range vmsOfNodeGroups {
+		deploymentsInfo[nodeGroup] = buildDeployments(tfPluginClient, vms, nodeGroups[nodeGroup], sshKeys)
+	}
+	return deploymentsInfo
+}
+
+func massDeploy(tfPluginClient deployer.TFPluginClient, ctx context.Context, deployemnts groupDeploymentsInfo) ([]string, error) {
```
AbdelrahmanElawady: `deployemnts` typo

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	vmsInfo := loadDeploymentsInfo(tfPluginClient, deployemnts.deploymentsInfo)
+
+	return vmsInfo, nil
+}
+
+func buildDeployments(tfPluginClient deployer.TFPluginClient, vms []Vms, nodesIDs []int, sshKeys map[string]string) groupDeploymentsInfo {
+	var vmDeployments []*workloads.Deployment
+	var networkDeployments []*workloads.ZNet
+	var deployemntsInfo []vmDeploymentInfo
```
AbdelrahmanElawady: same here

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+	}
+	return groupDeploymentsInfo{vmDeployments: vmDeployments, networkDeployments: networkDeployments, deploymentsInfo: deployemntsInfo}
+}
+
+func loadDeploymentsInfo(tfPluginClient deployer.TFPluginClient, deployemnts []vmDeploymentInfo) []string {
+	vmsInfo := []string{}
+	for _, info := range deployemnts {
+		vm, err := tfPluginClient.State.LoadVMFromGrid(info.nodeID, info.vmName, info.deploymentName)
+		if err != nil {
```
AbdelrahmanElawady: can this be done concurrently? we don't need to wait on each vm. and as far as I know the node client supports concurrent requests.

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,283 @@
...
+
+func filterNodes(tfPluginClient deployer.TFPluginClient, groups []NodesGroup, ctx context.Context) (map[string][]int, map[string]error) {
+	failedGroups := map[string]error{}
+	filteredNodes := map[string][]int{}
+	for _, group := range groups {
+		filter := types.NodeFilter{}
+		statusUp := "up"
+		filter.Status = &statusUp
+		filter.TotalCRU = &group.FreeCPU
+		filter.FreeMRU = &group.FreeMRU
```
AbdelrahmanElawady: `filter.FreeMRU` is using bytes while `group.FreeMRU` is using MB, right?

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,283 @@
...
+	var lock sync.Mutex
+	var wg sync.WaitGroup
+
+	deploymentStart := time.Now()
+
+	for nodeGroup, deployemnts := range groupsDeployments {
+		wg.Add(1)
+		go func(group string, deployemnts groupDeploymentsInfo) {
+			defer wg.Done()
+			info, err := massDeploy(tfPluginClient, ctx, deployemnts)
```
AbdelrahmanElawady: also if I remember correctly, I think there is an issue with deploying concurrently as multiple requests to the chain with the same mnemonic is not possible but I am not sure.

mik-tf:
I think this work is amazing. I am testing tfrobot and it works great!

I find that the conf.yaml example doesn't work on my end. Here's what works:

```
node_groups:
  - name: group_a
    nodes_count: 3 # amount of nodes to be found
    free_cpu: 2 # number of logical cores
    free_mru: 16384 # amount of memory in MB
    free_ssd: 100 # amount of ssd storage in GB
    free_hdd: 50 # amount of hdd storage in GB
    dedicated: false # are nodes dedicated
    pubip4: false # should the nodes have free ip v4
    pubip6: false # should the nodes have free ip v6
    certified: false # should the nodes be certified(if false the nodes could be certified of diyed) 
    region: "europe" # region could be the name of the continents the nodes are located in (africa, americas, antarctic, antarctic ocean, asia, europe, oceania, polar)
vms:
  - name: examplevm
    vms_count: 5 # amount of vms with the same configurations
    node_group: group_a # the name of the predefined group of nodes
    cpu: 1 # number of logical cores
    mem: 256 # amount of memory in MB
    ssd: # list of ssd storage needed to be mounted to the vm
      - size: 5 # size in GB
        mount_point: /mnt/ssd
    pubip4: false
    pubip6: false
    flist: https://hub.grid.tf/tf-official-apps/base:latest.flist
    entry_point: /sbin/zinit init
    root_size: 0 # root size in GB
    ssh_key: example1 # the name of the predefined ssh key
ssh_keys: # map of ssh keys with key=name and value=the actual ssh key
  example1: ssh_key1
mnemonic: example-mnemonic # mnemonic of the user
network: dev # eg: main, test, qa, dev
```

Basically, there was a - mising in front of size:, and ssh_key: # map... should be ssh_keys: # map... (add an s to get the proper go type from types.go)

mik-tf:
@Eslam-Nawara Once again, amazing work.

Do you think it is possible to add commands from tfcmd into tfrobot?
For example, to get contracts, and to delete contracts too. 
Ideally, we could delete contracts in blocks. 

Otherwise, users will end up using both tools for their deployments (tfrobot for mass deployments, tfcmd to delete and get contracts, for example). Which isn't bad, but it's less interesting in terms of UX.

mass-deployer/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+COPY /grid-proxy .
+
+WORKDIR /src/mass-deployer
+
+COPY /mass-deployer/go.mod .
+
+RUN go mod download
+
+COPY /mass-deployer/ ./
```
xmonader: Why ./ instead of . ?

mass-deployer/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+4.  Create a new configuration file.
+
+You can use this [example](./example/conf.yaml) for guidance, and make sure to replace placeholders and adapt the groups based on your actual project details.
+
+5.  Run the deployer with path to the config file
+```bash
+$ tfrobot -c path/to/your/config.yaml
+```
+
```
xmonader, line 55: Include an example for the config please (don't put your mnemonic)

mass-deployer/pkg/mass-deployer/types.go
```diff
@@ -0,0 +1,44 @@
+package deployer
+
+type Config struct {
+	NodeGroups []NodesGroup      `yaml:"node_groups"`
+	Vms        []Vms             `yaml:"vms"`
+	SSHKeys    map[string]string `yaml:"ssh_keys"`
```
xmonader: is that a map to be able to reference sshkeys by name? if not maybe it's better to convert to slice

mass-deployer/pkg/mass-deployer/types.go
```diff
@@ -0,0 +1,44 @@
+package deployer
```
xmonader, line 1: also please check this package to integrate validators https://pkg.go.dev/gopkg.in/validator.v2

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,103 @@
...
+	"os"
+	"slices"
+	"strings"
+
+	"github.com/cosmos/go-bip39"
+	deployer "github.com/threefoldtech/tfgrid-sdk-go/mass-deployer/pkg/mass-deployer"
+	"gopkg.in/yaml.v3"
+)
+
+func ParseConfig(configFile []byte) (deployer.Config, error) {
```
xmonader: this should be accepting io.Reader not []byte

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,103 @@
...
+
+	err := yaml.Unmarshal(configFile, &conf)
+	if err != nil {
+		return deployer.Config{}, err
+	}
+
+	mnemonic := conf.Mnemonic
+	network := conf.Network
+
+	if strings.TrimSpace(mnemonic) == "" {
```
xmonader: https://pkg.go.dev/gopkg.in/validator.v2 nonzero validator can help alot to remove some of the code repeatition in the rest of the validations

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,103 @@
...
+		mnemonic = os.Getenv("MNEMONIC")
+	}
+
+	if strings.TrimSpace(network) == "" {
```
xmonader: can be abstracted to a function `getValue(val, envVar)`

network = getValueOrEnv(network, "NETWORK")

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,103 @@
...
+	err := yaml.Unmarshal(configFile, &conf)
+	if err != nil {
+		return deployer.Config{}, err
+	}
+
+	mnemonic := conf.Mnemonic
+	network := conf.Network
+
+	if strings.TrimSpace(mnemonic) == "" {
+		mnemonic = os.Getenv("MNEMONIC")
```
xmonader: env keys extract to private consts to avoid typos (lots of time ppl confuse mnemonic with mnemonics)

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,103 @@
...
+	if strings.TrimSpace(network) == "" {
+		network = os.Getenv("NETWORK")
+	}
+
+	if !bip39.IsMnemonicValid(mnemonic) {
+		return deployer.Config{}, fmt.Errorf("invalid user mnemonic: %s", mnemonic)
+	}
+	networks := []string{"dev", "test", "qa", "main"}
+	if !slices.Contains(networks, network) {
+		return deployer.Config{}, fmt.Errorf("invalid netwok: %s", network)
```
xmonader: netwok -> network and include viable options in the error message

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,103 @@
...
+		if strings.TrimSpace(vm.Name) == "" {
+			return deployer.Config{}, fmt.Errorf("vms group name is invalid, shouldn't be empty")
+		}
+		if vm.Count == 0 {
+			return deployer.Config{}, fmt.Errorf("vms_count in vms group: %s is invalid, shouldn't be equal to 0", vm.Name)
+		}
+		if !slices.Contains(nodeGroupsNames, strings.TrimSpace(vm.Nodegroup)) {
+			return deployer.Config{}, fmt.Errorf("invalid node_group: %s in vms group: %s", vm.Nodegroup, vm.Name)
+		}
+		if vm.FreeCPU == 0 {
```
xmonader: check terraform validation e.g max cpu is 32

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		for _, disk := range vm.SSDDisks {
+			if disk.Size <= 0 {
+				return deployer.Config{}, fmt.Errorf("ssd disk size in vms group: %s is invalid, shouldn't be equal to 0", vm.Name)
+			}
+			if strings.TrimSpace(disk.Mount) == "" {
+				return deployer.Config{}, fmt.Errorf("vms group %s mount point is invalid, shouldn't be empty", vm.Name)
+			}
+		}
+		if strings.TrimSpace(vm.Flist) == "" {
```
xmonader: should end with .flist or .fl (and in case of .flist you should be able to append .md5 to get its md5 and thus, valid)

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,103 @@
...
+			return deployer.Config{}, fmt.Errorf("vms group %s flist is invalid, shouldn't be empty", vm.Name)
+		}
+		if strings.TrimSpace(vm.Entrypoint) == "" {
+			return deployer.Config{}, fmt.Errorf("vms group %s entry_point is invalid, shouldn't be empty", vm.Name)
+		}
+		if _, ok := conf.SSHKeys[vm.SSHKey]; !ok {
+			return deployer.Config{}, fmt.Errorf("vms group %s ssh_key is invalid, should be valid name refers to one of ssh_keys list", vm.Name)
+		}
+	}
+
```
xmonader, line 66:
```
	if err := validateMnemonic(mnemonic); err != nil {
		return deployer.Config{}, err
	}
	if err := validateNetwork(network); err != nil {
		return deployer.Config{}, err
	}
	if err := validateNodeGroups(); err != nil {
		return deployer.Config{}, err
	}
	if err := validateVMs(); err != nil {
		return deployer.Config{}, err
    }

```
I suggest you change the code to be like the following with multiple inline functions to validate

mass-deployer/internal/parser/parser_test.go
```diff
@@ -0,0 +1,149 @@
...
+    ssh_key: example1
+ssh_keys: 
+  example1: ssh-key1
+network: dev
+mnemonic: rival oyster defense garbage fame disease mask mail family wire village vibrant index fuel dolphin
+    `
+		_, err := ParseConfig([]byte(conf))
+		assert.NoError(t, err)
+	})
+}
```
xmonader, line 290: Can be expanded with more descriptive test names
- single zone single vm
- single zone many vms
- no zones single vm
- empty configs
- zone with < vm required resources 
- wrong filters (wrong region, country, invalid farm, .. etc)
- special characters in names used (includugin spaces)
- check for root disk 
- flists cases (.flist, .fl) and .flist that doesn't work when you append .md5 to the url and empty one
- cases for sshkeys (in the integration tests we need to see if they get added correctly to the machines)
- cpu/ram limits

could be more

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,281 @@
...
+
+	if len(passedGroups) > 0 {
+		fmt.Println("ok:")
+	}
+	for group, info := range passedGroups {
+		fmt.Printf("%s: \n%+v\n", group, info)
+	}
+
+	if len(failedGroups) > 0 {
+		fmt.Println("error:")
```
xmonader: errors always go to stderr `fmt.Fprintln(os.Stderr, .....` and same for the failed groups

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,281 @@
...
+			SizeGB: int(disk.Size),
+		}
+
+		disksWorkloads = append(disksWorkloads, DiskWorkload)
+		mountsWorkloads = append(mountsWorkloads, workloads.Mount{DiskName: DiskWorkload.Name, MountPoint: disk.Mount})
+	}
+	return
+}
+
+func convertGBToBytes(gb uint64) *uint64 {
```
xmonader: Why pointer to uint64? it's either 0 or you got a value to pass no?

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,281 @@
...
+	bytes := gb * 1024 * 1024 * 1024
+	return &bytes
+}
+
+func convertMBToBytes(mb uint64) *uint64 {
```
xmonader: same as above

mass-deployer/pkg/mass-deployer/types.go
Eslam-Nawara: Yes, to reference sshkeys easily

grid-client/deployer/deployer.go
```diff
@@ -448,7 +447,7 @@ func (d *Deployer) BatchDeploy(ctx context.Context, deployments map[uint32][]gri
 	}
 
 	contracts, index, err := d.substrateConn.BatchCreateContract(d.identity, contractsData)
-	if err != nil && index == nil {
+	if err != nil {
```
AbdelrahmanElawady: Removing this check removes the whole purpose of this method as non-atomic deployer. Because if index point to a contract 5 for example that means 4 contracts were created but there deployments won't be created. which other deployers are using it based on that (some contracts and deployments will be created and other will fail). also it won't cancel the contracts created so it will leave some contracts behind without the user knowing.

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,281 @@
...
+func RunDeployer(cfg Config) error {
+	ctx := context.Background()
+
+	tfPluginClient, err := setup(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to create deployer: %v", err)
+	}
+
+	groupsNodes, failed := filterNodes(tfPluginClient, cfg.NodeGroups, ctx)
+	failedGroups := failed
```
AbdelrahmanElawady: will this work?
```suggestion
	groupsNodes, failedGroups := filterNodes(tfPluginClient, cfg.NodeGroups, ctx)
```

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,281 @@
...
+		nodesIDs := []int{}
+		for _, node := range nodes {
+			nodesIDs = append(nodesIDs, node.NodeID)
+		}
+		filteredNodes[group.Name] = nodesIDs
+	}
+	return filteredNodes, failedGroups
+}
+
+func parseVMs(tfPluginClient deployer.TFPluginClient, vms []Vms, nodeGroups map[string][]int, sshKeys map[string]string) map[string]groupDeploymentsInfo {
```
AbdelrahmanElawady: `tfPluginClient` is not used

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,281 @@
...
+func massDeploy(tfPluginClient deployer.TFPluginClient, ctx context.Context, deployments groupDeploymentsInfo) ([]string, error) {
+	err := tfPluginClient.NetworkDeployer.BatchDeploy(ctx, deployments.networkDeployments)
+	if err != nil {
+		return []string{}, err
+	}
+
+	err = tfPluginClient.DeploymentDeployer.BatchDeploy(ctx, deployments.vmDeployments)
+	if err != nil {
+		return []string{}, err
+	}
```
AbdelrahmanElawady: In case of errors, some deployments might be created here so we need to make sure before returning that anything created is destroyed.

mass-deployer/internal/parser/parser_test.go
Eslam-Nawara: do you mean by zone timezone here?
I added more tests for the limits and the flists
the other tests can be created in the deployer as the parser doesn't check if the vm resources actually exist

mass-deployer/pkg/mass-deployer/deployer.go
Eslam-Nawara: it's a pointer to make it easy [here](https://github.com/threefoldtech/tfgrid-sdk-go/pull/527/files#diff-c4ca62f5c85b79fde950680322cb4a2735539c7a947aa63471ce63ee07bacd27R95-R102) to use the returned value directly in the filter, instead of saving the returned value in a variable then using it

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return deployer.Config{}, fmt.Errorf("failed to read the config file")
+	}
+
+	err = yaml.Unmarshal(configFile, &conf)
+	if err != nil {
+		return deployer.Config{}, err
+	}
+
+	conf.Mnemonic, err = getEnvOrValue(conf.Mnemonic, mnemonicKey)
```
xmonader: It was getValueOrEnv, but now you changed the name and the parameters order is confusing

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,80 @@
...
+
+	if err := validateVMs(conf.Vms, nodeGroupsNames, conf.SSHKeys); err != nil {
+		return deployer.Config{}, err
+	}
+
+	log.Info().Msg("done validating configuration file")
+	return conf, nil
+}
+
+func getEnvOrValue(value, valueType string) (string, error) {
```
xmonader: wrong names should be value, envKey instead and make it getValueOrEnv so the order of the parameters are `the value` or `the env key`

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,80 @@
...
+	valueType = strings.ToUpper(valueType)
+	if strings.TrimSpace(value) == "" {
+		value = os.Getenv("valueType")
```
xmonader: and now this will be always empty, because you are asking of env variable `"valueType"`

mass-deployer/internal/parser/parser_test.go
```diff
@@ -0,0 +1,359 @@
...
+	"github.com/rs/zerolog/log"
+	"github.com/stretchr/testify/assert"
+	deployer "github.com/threefoldtech/tfgrid-sdk-go/mass-deployer/pkg/mass-deployer"
+	"gopkg.in/yaml.v3"
+)
+
+func TestParseConfig(t *testing.T) {
+	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
+	tempDir := t.TempDir()
+	configPath := path.Join(tempDir, "conig.yaml")
```
xmonader: config.yaml

mass-deployer/internal/parser/parser_test.go
```diff
@@ -0,0 +1,359 @@
...
+		Network:  "dev",
+	}
+	t.Run("invalid yaml format", func(t *testing.T) {
+		conf := ` {
+  "node_groups": [
+    {
+      "nodes_count": 10
+    }
+} `
+		err := os.WriteFile(configPath, []byte(conf), 0667)
```
xmonader: Why do you need to do it that way? you can just get a reader of the string and pass it to the ParseConfig function

mass-deployer/internal/parser/parser_test.go
```diff
@@ -0,0 +1,359 @@
...
+		assert.NoError(t, err)
+
+		configFile, err := os.Open(configPath)
+		assert.NoError(t, err)
+
+		_, err = ParseConfig(configFile)
+		assert.Error(t, err)
+	})
+
+	t.Run("memory exeed limit in vm", func(t *testing.T) {
```
xmonader: exceed

mass-deployer/internal/parser/parser_test.go
```diff
@@ -0,0 +1,359 @@
...
+		assert.NoError(t, err)
+
+		configFile, err := os.Open(configPath)
+		assert.NoError(t, err)
+
+		_, err = ParseConfig(configFile)
+		assert.Error(t, err)
+	})
+
+	t.Run("root size exeed limit in vm", func(t *testing.T) {
```
xmonader: exceed

mass-deployer/internal/parser/parser_test.go
xmonader: I meant group, sorry for the confusion

mass-deployer/pkg/mass-deployer/deployer.go
xmonader: The function's role should be converting to Bytes, and that's it and it shouldn't be concerned with how you will use it

mass-deployer/pkg/mass-deployer/types.go
```diff
@@ -0,0 +1,46 @@
...
+	Certified  bool   `yaml:"certified"`
+	Regions    string `yaml:"regions"`
+}
+
+type Vms struct {
+	Name       string `yaml:"name" validate:"nonzero"`
+	Count      uint64 `yaml:"vms_count" validate:"nonzero"`
+	Nodegroup  string `yaml:"node_group" validate:"nonzero"`
+	FreeCPU    uint64 `yaml:"cpu" validate:"nonzero,max=32"`
+	FreeMRU    uint64 `yaml:"mem" validate:"nonzero,min=256,max=262144"` // max 256 GB
```
xmonader: in the comment it's min not max

mass-deployer/pkg/mass-deployer/types.go
```diff
@@ -0,0 +1,46 @@
...
+	PublicIP6  bool   `yaml:"public_ip6"`
+	Planetary  bool   `yaml:"planetary"`
+	Flist      string `yaml:"flist" validate:"nonzero"`
+	Rootsize   uint64 `yaml:"root_size" validate:"max=10240"` // max 10 TB
+	Entrypoint string `yaml:"entry_point" validate:"nonzero"`
+	SSHKey     string `yaml:"ssh_key" validate:"nonzero"`
+}
+
+type Disk struct {
+	Size  uint64 `yaml:"size" validate:"nonzero"`
```
xmonader: I suggest you make that at least 15GB min in validation

xmonader (changes requested):

mass-deployer/pkg/mass-deployer/types.go
Eslam-Nawara: it's max 256GB, the min is 256MB, I'll add both

mass-deployer/cmd/root.go
```diff
@@ -0,0 +1,53 @@
...
+
+var rootCmd = &cobra.Command{
+	Use:   "tfrobot",
+	Short: "A tool for deploying groups of vms on Threefold Grid",
+
+	Run: func(cmd *cobra.Command, args []string) {
+		configPath, err := cmd.Flags().GetString("config")
+		if err != nil || configPath == "" {
+			log.Fatal().Err(err).Msg("error in config file")
+			return
```
AbdelrahmanElawady: no need for return here Fatal will exit

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,78 @@
...
+	networkKey  = "NETWORK"
+)
+
+func ParseConfig(file io.Reader) (deployer.Config, error) {
+	conf := deployer.Config{}
+	nodeGroupsNames := []string{}
+
+	configFile, err := io.ReadAll(file)
+	if err != nil {
+		return deployer.Config{}, fmt.Errorf("failed to read the config file")
```
AbdelrahmanElawady: we should include the err returned from ReadAll

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,78 @@
...
+
+	conf.Network, err = getValueOrEnv(conf.Network, networkKey)
+	if err != nil {
+		return deployer.Config{}, err
+	}
+
+	log.Info().Msg("validating configuration file")
+
+	if err := validator.Validate(conf); err != nil {
+		return deployer.Config{}, fmt.Errorf("failed to validate config: %+v", err)
```
AbdelrahmanElawady: errors should be formatted with %w to be wrapped

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,78 @@
...
+	if err := validateVMs(conf.Vms, nodeGroupsNames, conf.SSHKeys); err != nil {
+		return deployer.Config{}, err
+	}
+
+	log.Info().Msg("done validating configuration file")
+	return conf, nil
+}
+
+func getValueOrEnv(value, envKey string) (string, error) {
+	envKey = strings.ToUpper(envKey)
```
AbdelrahmanElawady, line 72: envKey is not used

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,78 @@
...
+	if strings.TrimSpace(value) == "" {
```
AbdelrahmanElawady: extra if here? I think one of them was supposed to be the env key thing?

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,78 @@
...
+	mnemonicKey = "MNEMONIC"
+	networkKey  = "NETWORK"
+)
+
+func ParseConfig(file io.Reader) (deployer.Config, error) {
+	conf := deployer.Config{}
+	nodeGroupsNames := []string{}
+
+	configFile, err := io.ReadAll(file)
+	if err != nil {
```
rawdaGastan, line 25: return the error too with the msg

mass-deployer/internal/parser/parser.go
rawdaGastan: `%w`

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	log.Info().Msg("done validating configuration file")
+	return conf, nil
+}
+
+func getValueOrEnv(value, envKey string) (string, error) {
+	envKey = strings.ToUpper(envKey)
+	if strings.TrimSpace(value) == "" {
+		if strings.TrimSpace(value) == "" {
```
rawdaGastan: repeated condition? also use `len`

mass-deployer/internal/parser/parser.go
```diff
@@ -0,0 +1,78 @@
...
+			return "", fmt.Errorf("couldn't find valid %s", envKey)
```
rawdaGastan: `could not`

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,244 @@
...
+
+	deploymentStart := time.Now()
+
+	for _, nodeGroup := range cfg.NodeGroups {
+		log.Info().Msgf("running deployer for node group %s", nodeGroup.Name)
+		b := retry.NewFibonacci(1 * time.Nanosecond)
+
+		firstTrial := true
+		if err := retry.Do(ctx, retry.WithMaxRetries(2, b), func(ctx context.Context) error {
```
AbdelrahmanElawady: should the retries be configurable from user?

mass-deployer/internal/parser/validators.go
```diff
@@ -0,0 +1,61 @@
...
+	"slices"
+	"strings"
+
+	"github.com/cosmos/go-bip39"
+	deployer "github.com/threefoldtech/tfgrid-sdk-go/mass-deployer/pkg/mass-deployer"
+)
+
+func validateMnemonic(mnemonic string) error {
+	if !bip39.IsMnemonicValid(mnemonic) {
+		return fmt.Errorf("invalid user mnemonic: %s", mnemonic)
```
rawdaGastan: remove `user`

mass-deployer/internal/parser/validators.go
```diff
@@ -0,0 +1,61 @@
...
+	return nil
+}
+
+func validateVMs(vms []deployer.Vms, nodeGroups []string, sskKeys map[string]string) error {
+	for _, vm := range vms {
+		if !slices.Contains(nodeGroups, strings.TrimSpace(vm.Nodegroup)) {
+			return fmt.Errorf("invalid node_group: %s in vms group: %s", vm.Nodegroup, vm.Name)
+		}
+		if _, ok := sskKeys[vm.SSHKey]; !ok {
+			return fmt.Errorf("vms group %s ssh_key is invalid, should be valid name refers to one of ssh_keys list", vm.Name)
```
rawdaGastan: `list` -> `map`

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,243 @@
...
+
+			vmDeployments = append(vmDeployments, &deployment)
+			networkDeployments = append(networkDeployments, &network)
+			deploymentsInfo = append(deploymentsInfo, vmDeploymentInfo{nodeID: nodeID, deploymentName: deployment.Name, vmName: w.Name})
+		}
+	}
+	return groupDeploymentsInfo{vmDeployments: vmDeployments, networkDeployments: networkDeployments, deploymentsInfo: deploymentsInfo}
+}
+
+func cancelContractsOfFailedDeployments(tfPluginClient deployer.TFPluginClient, networkDeployments []*workloads.ZNet, vmDeployments []*workloads.Deployment) {
```
AbdelrahmanElawady, line 170: this function actually cancels working deployments as the contract won't be set in the workloads unless it managed to deploy fine. and using after failed batch deployments make the whole process of deploying a batch of networks and vms start from the beginning instead of retying only the failed ones.

rawdaGastan (changes requested):

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,243 @@
...
+	"gopkg.in/yaml.v3"
+
+	"github.com/rs/zerolog/log"
+	"github.com/threefoldtech/tfgrid-sdk-go/grid-client/deployer"
+	"github.com/threefoldtech/tfgrid-sdk-go/grid-client/workloads"
+	"github.com/threefoldtech/zos/pkg/gridtypes"
+)
+
+func RunDeployer(cfg Config) error {
+	ctx := context.Background()
```
AbdelrahmanElawady: maybe take a context from the caller where the user can specify a timeout?

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -0,0 +1,243 @@
...
+		return fmt.Errorf("failed to create deployer: %v", err)
+	}
+
+	deploymentStart := time.Now()
+
+	for _, nodeGroup := range cfg.NodeGroups {
+		log.Info().Msgf("running deployer for node group %s", nodeGroup.Name)
+		firstTrial := true
+
+		if err := retry.Do(ctx, retry.WithMaxRetries(2, retry.NewConstant(1*time.Nanosecond)), func(ctx context.Context) error {
```
AbdelrahmanElawady, line 35: we should maybe consider retrying only failed deployments not the whole group from the start. will save a lot of time I think

AbdelrahmanElawady (commented):
nice work

rawdaGastan:
Remaining work will be in another pr

rawdaGastan (approved):

Issue #265, opened 2023-08-01
long response time in some gridproxy requests


Omarabdul3ziz:
## Describe the bug

some requests to gridproxy takes long time to respond. the issue was found on mainnet but also can be reproduces on some requests on other networks

## To Reproduce
original issue was in requesting https://gridproxy.grid.tf/nodes?free_mru=8589934592&free_sru=55834574848&country=United%20States&available_for=1061&status=up

## Notes
- the issue happens when using the two filters `country` & `available_for` together.
- but it looks that `available_for` with `country_contains` filter works fine.
- there is also some broken requests like https://gridproxy.grid.tf/contracts?state=Deleted&type=rent and https://gridproxy.dev.grid.tf/contracts
- maybe the problem with `/contracts `endpoint is the subquery [here](https://github.com/threefoldtech/tfgrid-sdk-go/blob/c3044cec66862d5c86f4cdfde0f1c59c5bab6731/grid-proxy/internal/explorer/db/postgres.go#L703) but not sure if it the problem with `/nodes` endpoint also

Omarabdul3ziz:
the `/contracts` endpoint performance is fixed with this pr https://github.com/threefoldtech/tfgrid-sdk-go/pull/356

Omarabdul3ziz:
I can't find any particular problem with this request. also, it can't reproduce even on a large db mock.
hope it will be fixed with the improvements from this issue https://github.com/threefoldtech/tfgrid-sdk-go/issues/308, will update the progress there.

Omarabdul3ziz:
this https://github.com/threefoldtech/tfgrid-sdk-go/pull/533 should fixed the problem released with tag `v0.13.0` and deployed now on devnet

khaledyoussef24:
the issue is not reproducible any more endpoints work fine now
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/0cb29b14-e167-4987-9b1b-c07dd876e8da)

PR #533, opened 2023-12-10
optimize database queries 

Omarabdul3ziz:
### Description and changes

- denormalized some tables that have the most traffic of requests into `resources` and `public_ips` cache tables.
- added triggers for all the events that require updating these tables. and function to handle the update.
- update all the queries to use the new denormalized tables
- refactor the generate data script to a separate module `crafter` to be more customizable and add more methods for data modification/deletion to fire the triggers while needed
- change the flow on tests to run the crafter first to make sure it fires the triggers to update the cache tables. and then run the normal tests
- runs the test in parallel to minimize the time spent on testing
- updated the schema with newly generated one from processor-db on devnet
- handle some invalid data while testing against dev net dump
- add benchmark 

### Related issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/308
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/265
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/555
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/556


grid-proxy/internal/explorer/db/postgres.go
```diff
@@ -31,8 +31,107 @@ var (
 	ErrContractNotFound = errors.New("contract not found")
 )
 
+var views = []string{
```
muhamadazmy: This can instead by done with https://pkg.go.dev/embed package
this way the view can live in a separate file (say view.sql) and is embedded in the code so u can use it

muhamadazmy (changes requested):

xmonader (approved):

Issue #536, opened 2023-12-11
Add extra checks for nodes being filtered


AbdelrahmanElawady:
## Description

Filtering nodes with the grid-client in either `deployer.FilterNodes` or `deployer.GetPublicNode` will sometime return nodes that have status "up" but not actually responsive. We need to ping the nodes as an extra check to make sure the nodes are actually up.

related: #531

rawdaGastan:
Verified!

I tried to deploy a Kubernetes cluster and 2 VMs using a network with a WireGuard access, and it worked fine after the fix:
- now filter returns only working nodes

Issue #531, opened 2023-12-10
k8s deployment with WGAccess failed with `could not list node interfaces`


Eslam-Nawara:
## Describe the bug

[this](https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/grid-client/integration_tests/batch_k8s_test.go) integration test fails when WGAcess is required and for some reason it reports `failed to get node 11 endpoint` within the error message even when the deployment is not run using node 11 at all 

## Network
devnet

## Screenshots

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/67752395/7df73218-e330-449d-a7ff-0b6cde11aba3)

rawdaGastan:
Verification steps:

- Deploy a Kubernetes cluster using a network with a WireGuard access
- Deployed 2 VMs using a network with a WireGuard access

It worked fine after the fix:
- https://github.com/threefoldtech/tfgrid-sdk-go/pull/537

PR #537, opened 2023-12-11
Ping nodes while filtering

AbdelrahmanElawady:
### Description

Filtering nodes now will add extra check by pinging them.

### Related Issues

- #536 
- #531 

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation


xmonader (commented):
We should timeout these nodes if they don't reply within 2 seconds I think?

AbdelrahmanElawady:
@xmonader well we are using the context provided to the filter function, shouldn't it be the responsibility of the caller to set the timeout they want?

xmonader:
> @xmonader well we are using the context provided to the filter function, shouldn't it be the responsibility of the caller to set the timeout they want?

You're right 100%

xmonader (approved):

Issue #492, opened 2023-11-16
Allow FilterNodes to have a limit argument


Eslam-Nawara:
## Is your feature request related to a problem? Please describe

[FilterNode](https://github.com/threefoldtech/tfgrid-sdk-go/blob/b1ac1f1e83635d8ebe4a704892aa16587d71261f/grid-client/deployer/node_filter.go#L21) has default maximum number of nodes as 50, I think it would be much better to have a limit as a function argument to be able to filter more than 50 nodes

AbdelrahmanElawady:
Verified and tested with default limit vs 10 nodes:
```go
func main() {
	tf, err := deployer.NewTFPluginClient(mnemonic, "sr25519", network, "", "", "", 0, true)
	if err != nil {
		panic(err)
	}
	nodes, err := deployer.FilterNodes(context.Background(), tf, types.NodeFilter{Status: &status}, nil, nil, nil)
	if err != nil {
		panic(err)
	}
	fmt.Println(len(nodes))
	// 10 nodes max
	nodes, err = deployer.FilterNodes(context.Background(), tf, types.NodeFilter{Status: &status}, nil, nil, nil, 10)
	if err != nil {
		panic(err)
	}
	fmt.Println(len(nodes))
}
```
The logs show it tested 50 nodes in the first call and only 10 nodes in the second call, however it only returned 35 valid nodes in the first case and 4 in the second case. I think a clarification for the new option that it is the maximum number of nodes that will be tested might be better. However, it is working as expected.
Output:
```console
$ go run main.go
2:29PM DBG connecting url=wss://tfchain.grid.tf/ws
2023/12/13 14:29:50 Connecting to wss://tfchain.grid.tf/ws...
2:29PM INF starting peer session=tf-450527 twin=8209
2:29PM INF twin relay/public key didn't match, updating on chain ...
2:30PM DBG failed to ping node 260 error="context deadline exceeded"
2:30PM DBG failed to ping node 5 error="context deadline exceeded"
2:30PM DBG failed to ping node 303 error="context deadline exceeded"
2:30PM DBG failed to ping node 316 error="context deadline exceeded"
2:30PM DBG failed to ping node 39 error="context deadline exceeded"
2:30PM DBG failed to ping node 36 error="context deadline exceeded"
2:30PM DBG failed to ping node 315 error="context deadline exceeded"
2:30PM DBG failed to ping node 2 error="context deadline exceeded"
[...]

PR #557, opened 2023-12-17
fix NodeFilter limit option

Eslam-Nawara:
### Description

limit option doesn't ensure that the returned number of nodes are exactly equal to limit
this pr makes sure that the returned nodes are equals the requested limit or else it returns error

### Changes

- request more than one page of nodes with the requested filter with the same limit to ensure that enough nodes are returned 

### Related Issues

- #492


grid-client/deployer/node_filter.go
```diff
@@ -20,18 +20,69 @@ import (
...
-		twinID := uint64(tfPlugin.TwinID)
+		twinID := uint64(tf.TwinID)
 		options.AvailableFor = &twinID
 	}
 
+	if len(optionalLimit) == 0 {
+		return filterNodes(ctx, tf, options, ssdDisks, hddDisks, rootfs, limit)
+	}
+
+	pagesCount := 2
```
rawdaGastan: why you set it with 2?

grid-client/deployer/node_filter.go
```diff
@@ -20,18 +20,69 @@ import (
...
+	cnt := optionalLimit[0]
+	limit.Size = uint64(cnt)
+
+	if cnt < 100 {
```
rawdaGastan: I can't understand the condition and the pages count

grid-client/deployer/node_filter.go
```diff
@@ -20,18 +20,69 @@ import (
...
+		if err != nil {
+			log.Debug().Msg(err.Error())
+		}
+		return []types.Node{}, errors.Errorf("could not find enough node with options: %s", options)
+	}
+
+	return nodes[:cnt], allErrors
+}
+
+func filterNodes(ctx context.Context, tfPlugin TFPluginClient, options types.NodeFilter, ssdDisks, hddDisks []uint64, rootfs []uint64, limit types.Limit) ([]types.Node, error) {
```
rawdaGastan: rename it because I was confused when I saw it

grid-client/deployer/node_filter.go
```diff
@@ -20,18 +20,69 @@ import (
...
+		}(i)
+	}
+	wg.Wait()
+
+	if len(nodes) < cnt {
+		options, err := serializeOptions(options)
+		if err != nil {
+			log.Debug().Msg(err.Error())
+		}
+		return []types.Node{}, errors.Errorf("could not find enough node with options: %s", options)
```
rawdaGastan: `enough node` -> `enough nodes`

grid-client/deployer/node_filter.go
```diff
@@ -20,18 +20,69 @@ import (
...
+	if cnt < 100 {
+		pagesCount = 3
+	}
+
+	var nodes []types.Node
+	var allErrors error
+	var wg sync.WaitGroup
+	var lock sync.Mutex
+
+	for i := 0; i < pagesCount; i++ {
```
rawdaGastan: I think we should loop if we didn't reach the count and break if we finished the pages

rawdaGastan (changes requested):

grid-client/deployer/node_filter.go
```diff
@@ -19,29 +20,101 @@ import (
...
+	var lock sync.Mutex
+
+	nodesCount := optionalLimit[0]
+	limit := types.Limit{Randomize: true, Size: uint64(nodesCount), RetCount: true}
+
+	_, totalNodesCount, err := tf.GridProxyClient.Nodes(ctx, options, limit)
 	if err != nil {
 		return []types.Node{}, errors.Wrap(err, "could not fetch nodes from the rmb proxy")
 	}
+	totalPagesCount := int(math.Ceil(float64(totalNodesCount) / float64(limit.Size)))
```
rawdaGastan: I think we cannot calculate pages manually

grid-client/deployer/node_filter.go
```diff
@@ -19,29 +20,101 @@ import (
...
+				} else {
+					errs = errors.Wrap(errs, err.Error())
+				}
+				return
+			}
+
+			lock.Lock()
+			defer lock.Unlock()
+
+			if len(nodes) >= nodesCount {
```
rawdaGastan: why checking the same condition 2 times?

grid-client/deployer/node_filter.go
```diff
@@ -19,29 +20,101 @@ import (
...
+				return
+			}
+
+			nodesFounds, err := getNodes(ctx, tf, options, ssdDisks, hddDisks, rootfs, limit)
+			if err != nil {
+				lock.Lock()
+				defer lock.Unlock()
+
+				if errs == nil {
+					errs = err
```
rawdaGastan: Can't understand the `errs` behavior

rawdaGastan (changes requested):

rawdaGastan (approved):

PR #587, opened 2023-12-28
always power on dedicated nodes in farm (if they have a rent contract)

xmonader:
update canShutDown function to exclude if the node is dedicated condtion, and hook into shouldWakeup function to always return true in case of dedicdated nodes to be powered on as soon as the node is known to be dedicated

### Description

will make sure any node with rent contract to be powered on as soon as possible

### Changes

### Related Issues

- #585 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/farmerbot/internal/farmerbot.go
+++ b/farmerbot/internal/farmerbot.go
@@ -367,6 +367,9 @@ func (f *FarmerBot) shouldWakeUp(sub Substrate, node node, roundStart time.Time)
 	if node.powerState != off {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


xmonader:
maybe i got confused between n.dedicated and if the node has rentContracts active, will keep it as draft anyways until this is revisited

rawdaGastan:
Implemented here: https://github.com/threefoldtech/tfgrid-sdk-go/pull/596

Issue #680, opened 2024-01-29
Improve network wg port assignment


AbdelrahmanElawady:
## Description

Currently the grid-client assigns wireguard ports randomly to the networks from the range 2000 to 7999 leaving a lot of ports unused and increases the possibility of collision with other networks being deployed.
The range should be from 1024 to 32767 leaving privileged ports (0 to 1023) and ephemeral ports according to zos kernel (32768 to 60999).

```console
$ sysctl net.ipv4.ip_local_port_range
net.ipv4.ip_local_port_range = 32768  60999
```

Eslam-Nawara:
Discussed the issue with @AhmedHanafy725, and we agreed to leave the handling of the display of the output of the `iperf` tests to the client side and the test will still return the same output.

We also discussed if a public node testing the speed of itself would return a non-reasonable result, so I'm currently investigating that

While investigation, I noticed that the pref test on all `mainnet` nodes returns zero values for all fields with errors 
- node_id: 1
- twin_id: 7

- `error:unable to connect to server - server may have stopped running or use a different port, firewall issue, etc.: Operation timed out`
- `error:unable to connect to server - server may have stopped running or use a different port, firewall issue, etc.: Connection refused `

Issue #675, opened 2024-01-28
Set WG ports as part of the client state


AbdelrahmanElawady:
## Description

When deploying a batch of networks some networks assign the same WG port because they query the node for free ports, assign a random one and then wait until all networks do the same thing. Since it is not recorded anywhere which ports are used before deploying, sometimes multiple networks use the same ports.

Eslam-Nawara:
Verified: 
tried deploying batch of 20 networks all with wg access and now all networks get deployed with different wg ports
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/67752395/80bda84e-27c7-411b-9fbb-c798648cce81)

PR #681, opened 2024-01-29
Increase port range and store used ports

AbdelrahmanElawady:
### Description

Increase network port range and store used ports between different networks preprocessing.

### Changes

- network wireguard port can be from 1024 to 32767
- batch deploy network won't assign the same wireguard port

### Related Issues

- #680 
- #675 

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/grid-client/deployer/network_deployer.go
+++ b/grid-client/deployer/network_deployer.go
@@ -50,8 +50,9 @@ func (d *NetworkDeployer) Validate(ctx context.Context, znet *workloads.ZNet) er
 	return d.InvalidateBrokenAttributes(znet)
 }
 
-// GenerateVersionlessDeployments generates deployments for network deployer without versions
-func (d *NetworkDeployer) GenerateVersionlessDeployments(ctx context.Context, znet *workloads.ZNet) (map[uint32]gridtypes.Deployment, error) {
+// GenerateVersionlessDeployments generates deployments for network deployer without versions.
+// usedPorts can be used to exclude some ports from being assigned to networks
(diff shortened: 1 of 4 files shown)


grid-client/deployer/network_deployer.go
```diff
@@ -274,7 +275,7 @@ func (d *NetworkDeployer) Deploy(ctx context.Context, znet *workloads.ZNet) erro
 		return err
 	}
 
-	newDeployments, err := d.GenerateVersionlessDeployments(ctx, znet)
+	newDeployments, err := d.GenerateVersionlessDeployments(ctx, znet, nil)
```
rawdaGastan, line 278: What if I deployed some networks sequentially, how can I track the used ports?

AbdelrahmanElawady: we could add it to both Deploy and BatchDeploy but wouldn't the user just use BatchDeploy if they are deploying multiple networks?

rawdaGastan: Would it be better if we saved it in the network state?

AbdelrahmanElawady: I don't think we need store this data between deployments as the ports data are fetched from the node itself. so after deploying a network, the second time we deploy a network the ports of the first network will be fetched form the node itself (among other ports). so I don't see it needed outside of deploying a batch of networks.

rawdaGastan (approved):

Issue #664, opened 2024-01-23
show more info for nodes 


Omarabdul3ziz:
### Is your feature request related to a problem? Please describe

it is needed from the dashboard to view extra node info like:
- network speed (download/upload)
- dmi data

### Describe the solution you'd like

- get dmi data for each node from `zos.system.dmi` rmb call
- get speed data either from `zos.system.diagnostics` or `zos.perf.get(network)` calls

Omarabdul3ziz:
update: (wip)
- merged the indexer to be in a single package
- used a single rmb client across all the proxy
- introducing dmi watcher

Omarabdul3ziz:
update: (work is done)
- description & changes on the pr description https://github.com/threefoldtech/tfgrid-sdk-go/pull/726

Issue #636, opened 2024-01-16
🐞 [Bug]: [mainnet] inconsistency between nodes has_gpu filter and nodes non_gpu field 


khaledyoussef24:
### What happened?

version : 2.2.0
could not get gpus on mainnet all nodes are with 0 gpu in the dashboard

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/44776512-cffa-4157-850b-77df64fb6649)

after investigation using the gridproxy
when using <node_id>/gpu  :

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/03cfc88e-4b2a-4df3-8017-1d9830b94bae)

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/110984055/dfbc95a1-944b-4f7c-be97-8fac4e67b3c2)

 it always gives
"node_twin_id": 0,
and it gives 
"num_gpu": 0,

### which network/s did you face the problem on?

Main

### Relevant log output

```console
"node_twin_id": 0,
and it gives 
"num_gpu": 0,
```

A-Harby:
Also, the endpoint `https://gridproxy.dev.grid.tf/farms?node_has_gpu=true` doesn't list any farms on devnet, while there is a farm on devnet with a GPU (Farm QualiaDev with ID 4416).

Omarabdul3ziz:
to summarise the issue we are facing with GPUs now:
- [ ] filter `has_gpu=true` returns nodes with `num_gpu: 0`
- [ ] `/farms?node_has_gpu=true` doesn't return all farms returned in `/nodes?has_gpu=true`
- [ ] on `/nodes/<node_id>/gpu` endpoint response. `node_twin_id` field is always 0
- [ ] `num_gpu` in some nodes isn't equal to the length of the same node `/gpu` endpoint response

Omarabdul3ziz:
## Debugging findings 
> 
> * [ ]  filter `has_gpu=true` returns nodes with `num_gpu: 0`
> * [ ]  `/farms?node_has_gpu=true` doesn't return all farms returned in `/nodes?has_gpu=true`

I couldn't find the root cause for these two issues, but I suspect that it may be because of the status filter problem we had.
the GPU indexer to filter the up nodes before calling and listing its GPUs, which could lead to wrong data. 
However, both issues are no longer present on devnet or ninja stack. 
I verified this by:
[...]

PR #726, opened 2024-01-31
refactor the indexer

Omarabdul3ziz:
## Description & Changes
- combine indexers to be a single indexer with different watchers
- use a single RpcRmbClient across the proxy
- add more flags `-no-indexer` and `no-modify` to the commands & update the workflow to stop the indexer while integration tests
- add watchers for dmi/speed
- implement Scanner/Valuer for model structs
- add triggers on the new table to update the resources cache
- use `gorm` on the crafter for easier mock data generations for the custom model types
- add loader/tests for the new fields

## Related issues
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/664
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/636

## Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation


grid-proxy/cmds/proxy_server/main.go
```diff
@@ -57,9 +58,13 @@ type flags struct {
 	gpuIndexerResultWorkers     uint
 	gpuIndexerBatchWorkers      uint
 	maxPoolOpenConnections      int
-	// being 0 is helpful of making the data persistent while testing
-	healthIndexerWorkers  uint
-	healthIndexerInterval uint
+	healthIndexerWorkers        uint
+	healthIndexerInterval       uint
+	dmiWatcherWorkers           uint
```
xmonader: why did we move from the word "indexer" to watcher?

grid-proxy/pkg/types/nodes.go
```diff
@@ -53,6 +55,11 @@ type CapacityResult struct {
 	Used  Capacity `json:"used_resources"`
 }
 
+type Speed struct {
```
xmonader: I noticed there're multiple speeds for upload / download there's ipv4, ipv6 versions? which is it? and are

grid-proxy/internal/indexer/utils.go
```diff
@@ -0,0 +1,34 @@
+package indexer
+
```
xmonader, line 2: rename the file to filter_up_nodes.go

grid-proxy/internal/indexer/types.go
```diff
@@ -0,0 +1,47 @@
...
+// Property in section is in the form of key value pairs where values are optional
+// and may include a list of items as well.
+// k: [v]
+//
+//	[
+//		item1
+//		item2
+//		...
+//	]
+type DMI struct {
```
xmonader: Why didn't you bring the dmidecoder from zos ?

grid-proxy/internal/indexer/indexer.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/rs/zerolog/log"
+	"github.com/threefoldtech/tfgrid-sdk-go/rmb-sdk-go/peer"
+)
+
+const (
+	indexerCallTimeout = 10 * time.Second
```
xmonader: bump that to 30

grid-proxy/internal/indexer/indexer.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type Watcher interface {
```
xmonader: Rename to Indexer

grid-proxy/internal/indexer/indexer.go
```diff
@@ -0,0 +1,54 @@
...
+	Start(ctx context.Context)
+}
+
+type Indexer struct {
```
xmonader: Rename to Manager

grid-proxy/internal/indexer/indexer.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/rs/zerolog/log"
+	"github.com/threefoldtech/tfgrid-sdk-go/rmb-sdk-go/peer"
+)
+
+const (
+	indexerCallTimeout = 10 * time.Second
+)
+
+type Watcher interface {
+	Start(ctx context.Context)
```
xmonader: Should be Starter and Stopper and Pauser interfaces instead of Indexer or Watcher here

grid-proxy/internal/indexer/indexer.go
```diff
@@ -0,0 +1,54 @@
...
+		RmbClient: rmbClient,
+	}
+}
+
+func (i *Indexer) RegisterWatcher(name string, watcher Watcher) {
+	i.Watchers[name] = watcher
+}
+
+func (i *Indexer) Start() {
+	if i.Paused {
```
xmonader: It makes sense to resume after pausing, not sure why you return

grid-proxy/internal/indexer/indexer.go
```diff
@@ -0,0 +1,54 @@
...
+) *Indexer {
+	return &Indexer{
+		Watchers:  make(map[string]Watcher),
+		Paused:    paused,
+		Context:   ctx,
+		RmbClient: rmbClient,
+	}
+}
+
+func (i *Indexer) RegisterWatcher(name string, watcher Watcher) {
```
xmonader: RegisterIndexer or Register is enough
the 2nd param should be: indexer

grid-proxy/internal/indexer/indexer.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+type Indexer struct {
+	Watchers  map[string]Watcher
+	Paused    bool
+	Context   context.Context
+	RmbClient *peer.RpcClient
+}
+
+func NewIndexer(
```
xmonader: NewManager

grid-proxy/internal/indexer/indexer.go
```diff
@@ -0,0 +1,54 @@
...
+	ctx context.Context,
+	paused bool,
```
xmonader: Shouldn't be here, if for internal state, it should be defaulted to the zero value, false

grid-proxy/internal/indexer/gpu.go
```diff
@@ -1,19 +1,13 @@
...
-	"github.com/google/uuid"
 	"github.com/rs/zerolog/log"
-	substrate "github.com/threefoldtech/tfchain/clients/tfchain-client-go"
 	"github.com/threefoldtech/tfgrid-sdk-go/grid-proxy/internal/explorer/db"
 	"github.com/threefoldtech/tfgrid-sdk-go/grid-proxy/pkg/types"
 	"github.com/threefoldtech/tfgrid-sdk-go/rmb-sdk-go/peer"
-	rmbTypes "github.com/threefoldtech/tfgrid-sdk-go/rmb-sdk-go/peer/types"
 )
 
 const (
```
xmonader, line 13: prefix all of the consts with gpu (given there're multiple indexers in the same package)

grid-proxy/internal/indexer/gpu.go
```diff
@@ -37,16 +31,15 @@ type NodeGPUIndexer struct {
...
 	db db.Database,
 	indexerCheckIntervalMins,
 	batchSize,
 	nodesGPUResultsWorkers,
-	nodesGPUBufferWorkers uint) (*NodeGPUIndexer, error) {
-	indexer := &NodeGPUIndexer{
+	nodesGPUBufferWorkers uint) *NodeGPUIndexer {
+	return &NodeGPUIndexer{
 		db:                     db,
+		rpcClient:              rpcClient,
```
xmonader: is that client a reconnecting one, in long running processes you should be careful not to lose that single connection you have.

grid-proxy/internal/indexer/gpu.go
```diff
@@ -130,11 +108,25 @@ func (n *NodeGPUIndexer) runQueryGridNodes(ctx context.Context) {
 }
 
 func (n *NodeGPUIndexer) getNodeGPUInfo(ctx context.Context, nodeTwinID uint32) error {
-	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	subCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
xmonader: bump to 30

grid-proxy/internal/indexer/dmi.go
```diff
@@ -0,0 +1,176 @@
...
+				}
+				buffer = nil
+			}
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func parseDmiResponse(dmiResponse DMI) types.DmiInfo {
```
xmonader: replace everything related to parsing by zos dmidecode package, don't do it yourself

xmonader (changes requested):
I had to stop the review. 

1- Please make sure all of the indexers are having the similar interface 
2- Remove everything related to any manual parsing of the dmidecode
3- Prefix all of the constants related to each indexer with the indexer name e.g `healthIndexerNumWorkers`
4- for every indexer there should be
 - a `Start` method, can have its own interface `Starter`
 - a `Stop` method, can have its own interface `Stopper`
 - a `FindHealthyNodes` method
 - a `ResultsBatcher` method
 - a `DBBatchesUpserter` method
 
 All of them can be converted to functions whenever needed.
 
 Each indexer file should have the logic top to bottom 
 
 e.g 
 
- consts
- vars
- types
- factories
- entry point `Start` function
- then each of the methods asked for 

Note: consts and vars should be prefixed with the indexer name

Use the same terms everywhere: e.g status, state, numWorkers, workersCount, .. etc All of these need to be unified


in an upcoming improvement we can use singleflight to avoid expensive DB/network calls e.g all are asking for the healthy nodes within the same timeframe

Aside from that, the PR already has too many concerns, supporting the dmi, refactoring in the indexers, introducing more dependencies. Please in the future make sure it's scoped.

grid-proxy/pkg/types/nodes.go
Omarabdul3ziz: yes we have `ipv4` and `ipv6` for both `tcp` and `udp` these are four per an iperf server, 
for example devnet has two iperf servers on the two public nodes so we have total of `8` tests

a discussion was raised on the chat group and i remember the decision was to show the only non-zero value. should we do something different? or view them all?

grid-proxy/internal/indexer/utils.go
Omarabdul3ziz: it was planned to have more utils method for filtering up nodes/healthy nodes or call rmb

grid-proxy/internal/indexer/indexer.go
Omarabdul3ziz: actually `pause` was a misleading i need a flag to stop the indexer in this run, it is helpful when running integration tests to keep data unchanged from the indexer.

for a better naming i called the flag `-no-indexer` also made the decision for starting manager or not outside the pkg in the main file

grid-proxy/internal/indexer/gpu.go
Omarabdul3ziz: afaik yes, the rpcClient taking a substrate manager and handle the open/close connections when needed

grid-proxy/internal/indexer/dmi.go
Omarabdul3ziz: checked the dmi pkg in zos and the `DMI` response is already parsed from the pkg in zos.
what i am doing here is just selecting what we want to store instead of storing the whole dmi response

Omarabdul3ziz:
Update changes:
- configure a flag to stop or start the indexer manager on the startup. it is useful to use `-no-indexer` flag while running integration tests, to keep the loaded data in the mock client consistent with the actual data in DB.
- organize the database calls in `postgres.go` file by separating the indexer calls into a new file. also, save consistency for numeric `NodeTwinId` filed to be `uint32` in all methods
- remove the duplicated types across different pkgs and use only the ones in `pkg/types` pkgs db/tests/generator
- add new trigger on the `node_gpu` table, this trigger should be fired on insert/delete to update the `num_gpu` filed on the `resources_cache` table
- remove all implementation for `Scanner/Valuer` interfaces for jsonb fields in gorm model. instead easier and cleaner, use the registered json serializer. apply this for dmi fields/node power object
- modify the generation of `healthy_reports` mock data to use `gorm` functionality instead of creating a plain SQL string
- update and organize the schema file used on the db mock generation to be aligned with the new defined struct models.

Indexer pkg changes:
- import types directly from zos once needed instead of re-implement it, applied this for IperfTestResult and DMI
- create utils method, generic methods that can be used in all indexers to avoid repeating the code
- create a manager that register/control all the indexers
- define an `Indexer` interface that defines the needed methods for each indexer
- document the indexer package along with the registered indexers and their metadata
- implement `Indexer` interface for each registered interface, this requires some refactoring for the indexer especially the GPU indexer
- a better decision on old GPU node deletion by adding a timestamp expiration and check it is expired or not with each upsert, the allowed interval is the same as the check interval in the indexer an hour for now. this is better than checking the lastAdded twin id cause we have multiple caller/batcher that may conflict the results.
- add new trigger for the dmi indexer, to start querying the newly added nodes and don’t wait for the new rotation

grid-proxy/cmds/proxy_server/main.go
```diff
@@ -81,13 +86,18 @@ func main() {
...
+	flag.BoolVar(&f.noIndexer, "no-indexer", false, "do not start the indexer")
+	flag.UintVar(&f.indexerUpserterBatchSize, "indexer-upserter-batch-size", 20, "results batch size which collected before upserting")
+	flag.UintVar(&f.gpuIndexerIntervalMins, "gpu-indexer-interval", 60, "the interval that the GPU indexer will run")
+	flag.UintVar(&f.gpuIndexerNumWorkers, "gpu-indexer-workers", 100, "number of workers to process indexer GPU info")
+	flag.UintVar(&f.healthIndexerIntervalMins, "health-indexer-interval", 5, "node health check interval in min")
+	flag.UintVar(&f.healthIndexerNumWorkers, "health-indexer-workers", 100, "number of workers checking on node health")
+	flag.UintVar(&f.dmiIndexerIntervalMins, "dmi-indexer-interval", 60*24, "node dmi check interval in min")
+	flag.UintVar(&f.dmiIndexerNumWorkers, "dmi-indexer-workers", 1, "number of workers checking on node dmi")
+	flag.UintVar(&f.speedIndexerIntervalMins, "speed-indexer-interval", 5, "node speed check interval in min")
+	flag.UintVar(&f.speedIndexerNumWorkers, "speed-indexer-workers", 100, "number of workers checking on node speed")
```
muhamadazmy, line 101: Just wondering why do we need multiple indexer workers to query things from nodes, can't we have a single set of workers that they ask the node for the same thing for everything? 

I mean i see here, dmi-indexer-workers, gpu-indexer-workers, speed-indexer-workers (whatever that means), etc... 

This is just a question, not request for change

grid-proxy/internal/indexer/gpu.go
```diff
@@ -0,0 +1,177 @@
...
+		interval:        time.Duration(interval) * time.Minute,
+		nodeTwinIdsChan: make(chan uint32),
+		resultChan:      make(chan types.NodeGPU),
+		batchChan:       make(chan []types.NodeGPU),
+	}
+}
+
+func (n *NodeGPUIndexer) Start(ctx context.Context) {
+	go n.StartNodeFinder(ctx)
+	go n.startNodeTableWatcher(ctx)
```
muhamadazmy: Do these methods need to be public? I believe they are intended for private use by the indexer object. I suggest they are renamed to be private instaed

muhamadazmy: This applies to all other indexers here

grid-proxy/internal/indexer/health.go
```diff
@@ -0,0 +1,139 @@
...
+	rmbClient       *peer.RpcClient
+	nodeTwinIdsChan chan uint32
+	resultChan      chan types.HealthReport
+	batchChan       chan []types.HealthReport
+	indexerInterval time.Duration
+	indexerWorkers  uint
+	batchSize       uint
+}
+
+func NewNodeHealthIndexer(
```
muhamadazmy: I can see that the an indexer mainly does the same operation, I was wondering that an indexer logic can be abstracted more. 
From what i see an indexer has a set of options (say workers), he run indexing every certain amount of time. Then does some sort of batching to update the db. 
All this logic then can be unified. Then each indexer need to define 2 things.
- The function that does the "fetch" of the data. 
- The logic of "batch" db update.

Those 2 functions then can be part of an IndexerWork interface. And can has multiple implementation (say gpu and dmi) and then an Indexer instance can be created for each something like `NewIndexer(GPUIndexerWork())` for example. Then call `indexer.Start()`

muhamadazmy (changes requested):

muhamadazmy (approved):

Omarabdul3ziz:
Converting to draft to not merge till 3.13 reaches mainnet.

xmonader:
yalla beena

xmonader (approved):

Issue #715, opened 2024-01-31
Change mass deployer clean up strategy 


Eslam-Nawara:
### Is your feature request related to a problem? Please describe

Mass deployer clean up strategy is to remove all deployments of the failed group and rerun the deployer for that group (reported [here](https://github.com/threefoldtech/tfgrid-sdk-go/pull/527#discussion_r1469764638))

### Describe the solution you'd like

A [suggestion](https://github.com/threefoldtech/tfgrid-sdk-go/pull/527#discussion_r1469777337) is to update the deployer to rerun only the failed deployments

Eslam-Nawara:
WIP: 
- updated the retry strategy to only redeploy failed deployments.

- tried to use `ListContractsOfProjectName` to load deployed contracts from grid instead of keeping state of the deployments, but it doesn't list the contracts deployed just now for some reason, I would be required to wait for some time before I call it, not sure what the expected behavior is.

- sometimes the created contracts can't be loaded from state at all, with both methods of listing contracts, not sure what is the cause of it

the deployments exist on the playground with the same info

```console
5:30PM INF couldn't load from state error="failed to find deployment  on node 100: not found" deployment=examplevm4 node ID=100 vm=examplevm4
5:30PM INF couldn't load from state error="failed to find deployment  on node 150: not found" deployment=examplevm1 node ID=150 vm=examplevm1
5:30PM INF couldn't load from state error="failed to find deployment  on node 153: not found" deployment=examplevm3 node ID=153 vm=examplevm3
5:30PM INF couldn't load from state error="failed to find deployment  on node 146: not found" deployment=examplevm0 node ID=146 vm=examplevm0
5:30PM INF couldn't load from state error="failed to find deployment  on node 152: not found" deployment=examplevm2 node ID=152 vm=examplevm2
```

Eslam-Nawara:
### Blocked:
- continued working on updating the redeployment strategy
- fixed the `couldn't load from state` problem
- tested the new strategy and it works as expected

No we need to fix the `DeploymentDeployer.BatchDeploy` function as it suppose to be not atomic, but it still all deployments fails to deploy any deployment if one of the used networks failed, which is not the expected behavior reported in:
- #765

rawdaGastan:
Tried to deploy 100 VMs and some of them failed then the only failed deployments got deployed again and it worked

Issue #683, opened 2024-01-29
mass-deployer: continuation


Eslam-Nawara:
Add improvements and fix reported issues of the mass deployer

- [x] #690 
- [x] #688 
- [x] #689 
- [x] #687 
- [x] #686 
- [x] #684 
- [x] #685 
- [x] #694
- [x] #697
- [x] #700 
- [x] #715

Eslam-Nawara:
Allow users to configure retries

https://github.com/threefoldtech/tfgrid-sdk-go/pull/527#discussion_r1469727327

Eslam-Nawara:
Add mass deployer to sdk readme packages

xmonader:
convert to a checklist in the body of the issue instead of using comments for sub issues please

PR #743, opened 2024-02-04
update retry strategy

Eslam-Nawara:
### Description

update retry strategy to update only failed deployments and don't recreate the past ones

### Changes

run batch deploy for only deployments with 0 value for contract ID 

### Related Issues

- #715 
- #683 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -284,3 +293,56 @@ func parseDisks(name string, disks []Disk) (disksWorkloads []workloads.Disk, mou
...
 }
+
+func updateFailedDeployments(tfPluginClient deployer.TFPluginClient, nodesIDs []int, groupDeployments *groupDeploymentsInfo) {
+	// update nodes for failed networks and update vm deployments accordingly
+	// return if the networks were updated
+	var updated bool
+	for idx, deployment := range groupDeployments.networkDeployments {
+
+		nodeID := uint32(nodesIDs[idx%len(nodesIDs)])
+		if isFailedNetwork(*deployment) {
```
AbdelrahmanElawady: `isFailedNetwork` returns true if one contract is 0. can some contracts still exist with the 0 contract? if that's the case we will need to cancel these contracts too.

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -284,3 +293,56 @@ func parseDisks(name string, disks []Disk) (disksWorkloads []workloads.Disk, mou
...
+			groupDeployments.networkDeployments[idx].Nodes = []uint32{nodeID}
+		}
+	}
+
+	if updated {
+		return
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
AbdelrahmanElawady: it should get the ctx from the parent function

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -284,3 +293,56 @@ func parseDisks(name string, disks []Disk) (disksWorkloads []workloads.Disk, mou
...
+	for idx, deployment := range groupDeployments.vmDeployments {
+		if deployment.ContractID == 0 {
+			for _, contract := range groupDeployments.networkDeployments[idx].NodeDeploymentID {
+				if contract != 0 {
+					err := tfPluginClient.NetworkDeployer.Cancel(ctx, groupDeployments.networkDeployments[idx])
+					if err != nil {
+						log.Debug().Err(err)
+					}
+					break
+				}
```
AbdelrahmanElawady: we can cancel it directly and it will do nothing if it's not deployed I believe

mass-deployer/pkg/mass-deployer/deployer.go
Eslam-Nawara: it's not the case with mass deployer, the network has one node in every network

AbdelrahmanElawady: it still could deploy more than one in public node example

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -283,3 +293,38 @@ func parseDisks(name string, disks []Disk) (disksWorkloads []workloads.Disk, mou
...
+	for idx, deployment := range groupDeployments.vmDeployments {
+		if deployment.ContractID == 0 || isFailedNetwork(*groupDeployments.networkDeployments[idx]) {
+			nodeID := uint32(nodesIDs[idx%len(nodesIDs)])
+			groupDeployments.vmDeployments[idx].NodeID = nodeID
+			groupDeployments.networkDeployments[idx].Nodes = []uint32{nodeID}
+		}
+	}
+}
+
+func isFailedNetwork(network workloads.ZNet) bool {
```
AbdelrahmanElawady: there is no need for that now as failed networks should have no contracts

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -283,3 +293,38 @@ func parseDisks(name string, disks []Disk) (disksWorkloads []workloads.Disk, mou
...
+	if len(network.NodeDeploymentID) == 0 {
+		return true
+	} else {
```
rawdaGastan: remove else after return :D

rawdaGastan (changes requested):

AbdelrahmanElawady (approved):

rawdaGastan (approved):

rawdaGastan:
Good job 🔥 🔥

PR #772, opened 2024-02-06
buffered channel for incomingEnv and RWMutex

xmonader:
### Description

- Uses RWMutex
- Uses buffered channel for incomingEnv
### Changes
- Uses RWMutex
- Uses buffered channel for incomingEnv

## List of changes this PR includes
- Uses RWMutex
- Uses buffered channel for incomingEnv

### Related Issues

List of related issues

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/rmb-sdk-go/peer/examples/peer_pingmany/main.go
+++ b/rmb-sdk-go/peer/examples/peer_pingmany/main.go
@@ -1,0 +1,78 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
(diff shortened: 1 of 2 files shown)


muhamadazmy (approved):
This a really good fix
I think the main problem is that the channel with unbuffered, which made it block when write the response hence slow down all the calls to the route function.

The idea was that to `select` this with a default branch which means if no readers this should return immediately but it seems it is not the case in practice and it waits a little bit before it gives up

Issue #799, opened 2024-02-08
Allow twins caching in mass deployer


Eslam-Nawara:
### Is your feature request related to a problem? Please describe

rmb allows caching twins info for better performance

### Describe the solution you'd like

add new option to use rmb twin cache to improve deployer performance

Eslam-Nawara:
Work Completed:
- added a new flag `rmbInMemCache` to allow memory or disk cache
- increased default timeout from 10s to 60s
- updated all packages to use in memory cache
- updated mass deployer to use in disk cache

AbdelrahmanElawady:
Verified:

tfrobot still uses in memory cache, no cache can be found under `/tmp/rmb-cache` after running tfrobot.

```console
$ ls /tmp/rmb-cache
ls: cannot access '/tmp/rmb-cache': No such file or directory
```

Eslam-Nawara:
Work completed:
twin cache is now added to tfrobot 
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/67752395/b3a9be86-445a-4392-aa51-10b55963caa8)

AbdelrahmanElawady:
Verified:

`tfrobot` now uses `/tmp/rmb-cache` to cache twins:

```console
$ tree /tmp/rmb-cache 
/tmp/rmb-cache
└── tfchain.dev.grid.tf:443
    └── 192

1 directory, 1 file
```

PR #800, opened 2024-02-08
add rmb caching flag to `NewTFPluginClient`

Eslam-Nawara:
### Changes

add option to allow twins caching for 10 hours
increase client default timeout to 60s

### Related Issues

- #799 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/grid-client/deployer/tf_plugin_client.go
+++ b/grid-client/deployer/tf_plugin_client.go
@@ -209,15 +209,15 @@ func NewTFPluginClient(
 	ctx, cancel := context.WithCancel(context.Background())
 	tfPluginClient.cancelRelayContext = cancel
 
-	var peerOpts []peer.PeerOpt
+	 peerOpts := []peer.PeerOpt{
+	          peer.WithRelay(tfPluginClient.relayURL),
+	          peer.WithSession(sessionID)),
+	          peer.WithKeyType(keyType),
+	 }
+	 
 	if !rmbInMemCache {
 		peerOpts = append(peerOpts, peer.WithTwinCache(10*60*60)) // in seconds that's 10 hours
 	}
-
-	peerOpts = append(peerOpts, peer.WithRelay(tfPluginClient.relayURL))
-	peerOpts = append(peerOpts, peer.WithSession(sessionID))
-	peerOpts = append(peerOpts, peer.WithKeyType(keyType))
-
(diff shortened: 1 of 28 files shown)


grid-client/deployer/tf_plugin_client.go
```diff
@@ -105,6 +105,7 @@ func NewTFPluginClient(
 	proxyURL string,
 	rmbTimeout int,
 	showLogs bool,
+	allowRMBCache ...bool,
```
xmonader: should be just bool not ...bool, and should be rmbMemCache and if false we use the disk cache

grid-client/deployer/tf_plugin_client.go
```diff
@@ -199,16 +200,25 @@ func NewTFPluginClient(
...
-	rmbClient, err := peer.NewRpcClient(ctx, tfPluginClient.mnemonicOrSeed, manager, peer.WithRelay(tfPluginClient.relayURL), peer.WithSession(sessionID), peer.WithKeyType(keyType))
+	var peerOpts []peer.PeerOpt
+	if !rmbInMemCache {
+		peerOpts = append(peerOpts, peer.WithTwinCache(10*60*60)) // in seconds that's 10 hours
+	}
+
+	peerOpts = append(peerOpts, peer.WithRelay(tfPluginClient.relayURL))
+	peerOpts = append(peerOpts, peer.WithSession(sessionID))
+	peerOpts = append(peerOpts, peer.WithKeyType(keyType))
+
```
muhamadazmy: since you already have a fixed set of options you can easily do this instead of the above.
```suggestion
	 peerOpts := []peer.PeerOpt{
	          peer.WithRelay(tfPluginClient.relayURL),
	          peer.WithSession(sessionID)),
	          peer.WithKeyType(keyType),
	 }
	 
	if !rmbInMemCache {
		peerOpts = append(peerOpts, peer.WithTwinCache(10*60*60)) // in seconds that's 10 hours
	}
```

Also it's faster since the compiler can optimise this by already creating the array with the correct capacity

muhamadazmy (changes requested):

xmonader (approved):

muhamadazmy (approved):
looks good to me

PR #802, opened 2024-02-08
Add relay cache warmer

AbdelrahmanElawady:
### Description

Add Relay cache warmer to improve Relay performance.

### Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/899

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation

### DIFF
--- a/tools/relay-cache-warmer/graphql.go
+++ b/tools/relay-cache-warmer/graphql.go
@@ -9,6 +9,9 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
+
+	"github.com/gomodule/redigo/redis"
 )
 
 const query = `query TwinsQuery {
@@ -26,7 +29,10 @@ const query = `query TwinsQuery {
   }
 }`
 
-const twinsPerPage = 500
+const (
+	twinsPerPage     = 500
+	requestsInterval = 3 * time.Second
+)
 
 type Twin struct {
 	ID        uint      `json:"id"`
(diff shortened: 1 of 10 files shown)


tools/relay-cache-warmer/graphql.go
```diff
@@ -0,0 +1,190 @@
...
+func (u JSONArray) MarshalJSON() ([]byte, error) {
+	if u == nil || len(u) == 0 {
+		return []byte("null"), nil
+	}
+
+	var buf bytes.Buffer
+	_, err := buf.WriteRune('[')
+	if err != nil {
+		return nil, err
+	}
```
muhamadazmy: i think the buffer write should never fail (u can safely ignore the error) unless it failed to allocate memory which means your entire app will crash anyway.

If you check implementation for [WriteRune](https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/bytes/buffer.go;l=299) or even `Write` you will see it never returns an error

So in your code u can ignore the error or do something like 
```go
_, _ = buf.WriteRune('[') 
```
to make sure linters are happy

tools/relay-cache-warmer/main.go
```diff
@@ -0,0 +1,66 @@
...
+
+	var opts []redis.DialOption
+	if redisPassword != "" {
+		opts = append(opts, redis.DialPassword(redisPassword))
+	}
+
+	conn, err := redis.DialURL(redisURL, opts...)
+	if err != nil {
+		log.Fatalf("failed to connect to redis server: %s", err.Error())
+	}
```
muhamadazmy: Never use a single connection for a long running process because u can lose this connection at any moment in time. (what if redis was restarted) 

The right way to do this is by always using a Pool and ask the pool for a connection when it's needed then close it immediately (the connection)

the pool then can decide to keep the connection actually open and use it for something else or create a new one if this one was broken, etc...

tools/relay-cache-warmer/main.go
```diff
@@ -0,0 +1,66 @@
...
+
+	run(conn, graphql, time.Duration(interval)*time.Minute)
+
+}
+
+func run(conn redis.Conn, graphql string, interval time.Duration) {
+	for {
+		select {
+		case <-time.After(interval):
+			go func() {
```
muhamadazmy: Why does this run in a go route. Doesn't sound right!

Your loop should instead, does the synchronization of ALL twins (page by page) if this took 2 minutes it's then wait 8 minutes (our of 10) before it starts again. If it took instead 11 minutes then it can start immediately but shouldn't start mulitple routines otherwise u can have multiple routines trying to synchrnoize the same set of twins in parallel (potential collision also)

tools/relay-cache-warmer/graphql.go
```diff
@@ -0,0 +1,190 @@
...
+	_, err = buf.WriteRune(']')
+	if err != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+
+}
+
+func getTwins(graphql string) ([]Twin, error) {
```
muhamadazmy: The point of using pagination is that you can process the data in batches. It is not right to use pagination to still collect ALL twins in one list !! then save them all

Instead this should work as:
- get first batch (say 200 twin)
- save 200 twin
- take a very short break 
- get next 200
- save 
- and so on!

muhamadazmy (changes requested):

xmonader (approved):

xmonader (approved):

Issue #805, opened 2024-02-11
🐞 [Bug] TFRobot: Node groups name conflict


Eslam-Nawara:
### What happened?

When running the deployer multiple times with the same node group names, if the first deployment passed and the following deployment failed, while cancelation the deployer delete both the deployments of the first and the second deployments as both has the same node group name.

### which network/s did you face the problem on?

Main

### Relevant log output

```console
--
```

A-Harby:
Verified, Dev branch.

[TC2624 - Deploying 2 VMs group with same node group](https://app.testlodge.com/a/26076/projects/40893/suites/236972?expand_section=427886#case_4220274)

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/cd8b0e8b-e28b-49ef-81f4-7fd0d0787d1d)

PR #807, opened 2024-02-11
Add node groups name validation

Eslam-Nawara:
### Description

- Add node group validation to prevent conflicts of multiple node groups with the same name across deployments 

### Changes

- Add unique validation to node group names in the same config file
- Check if any previous deployments with the same mnemonics has the same node group name

### Related Issues

- #805 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


mass-deployer/internal/parser/validators.go
```diff
@@ -29,7 +33,51 @@ func validateNetwork(network string) error {
...
+
+	var multiErr error
+	var wg sync.WaitGroup
+	var lock sync.Mutex
+
+	for _, group := range nodeGroups {
+		wg.Add(1)
+		go func(group massDeployer.NodesGroup) {
+			defer wg.Done()
+			nodeGroupName := strings.TrimSpace(group.Name)
```
xmonader, line 46: extract the logic for the validation of a nodeGroup into a local scoped function or even a function

mass-deployer/internal/parser/validators.go
```diff
@@ -29,7 +33,51 @@ func validateNetwork(network string) error {
...
+		return err
+	}
+
+	var multiErr error
+	var wg sync.WaitGroup
+	var lock sync.Mutex
+
+	for _, group := range nodeGroups {
+		wg.Add(1)
+		go func(group massDeployer.NodesGroup) {
```
xmonader: I believe this should be refactored to use errgroup and fail on the first error, don't believe we need to find all of the issues, one of them is just enough.  https://pkg.go.dev/golang.org/x/sync/errgroup

mass-deployer/pkg/mass-deployer/deployer.go
```diff
@@ -20,8 +20,10 @@ import (
 	"github.com/threefoldtech/zos/pkg/gridtypes"
 )
 
-const maxDeploymentRetries = 5
-const maxGoroutinesToFetchState = 100
+const (
+	maxDeploymentRetries      = 5
```
xmonader: rename to DefaultMaxRetries and use that everywhere when needed if not overridden

xmonader (approved):

PR #816, opened 2024-02-12
Add cache-warmer to image publish workflow

AbdelrahmanElawady:
### Description

Add cache-warmer to image publish workflow

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -32,6 +32,8 @@ jobs:
             image: "tffarmerbot"
           - dir: "grid-proxy"
             image: "tfgridproxy"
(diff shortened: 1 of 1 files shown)


Omarabdul3ziz (approved):

.github/workflows/publish-images.yml
```diff
@@ -32,6 +32,8 @@ jobs:
             image: "tffarmerbot"
           - dir: "grid-proxy"
             image: "tfgridproxy"
+          - dir: "tools/relay-cache-warmer"
+            image: "tfcachewarmer"
```
muhamadazmy: I think a proper name is relay-cache-warmer otherwise the name is confusing

Issue #804, opened 2024-02-08
🐞 [Bug]: Farmerbot wake up threshold works on simple sum of all resource types


scottyeager:
### What happened?

The Farmerbot is currently calculating the resources used in the farm like this:

https://github.com/threefoldtech/tfgrid-sdk-go/blob/3bc870947dc8d0b4ff4f2687d65543928aac366f/farmerbot/internal/power.go#L140

This approach doesn't work:

* Summing together vCPU counts (CRU) and byte counts of RAM/disk (MRU/SRU/HRU) just doesn't make sense
* Nodes can have all their RAM reserved, or have their CPU heavily oversubscribed and never reach threshold

I suggest something like this:

* Evaluate threshold on CRU/MRU/SRU individually (wake another node if any one of these quantities exceeds the threshold percentage individually)
* Just ignore HRU for now (doing this right would involve adding some logic around whether there's a node available to wake up that has HRUs. Since basically nobody is reserving HDD on the grid today, we can figure it out later)

A-Harby:
Verified, Devnet.

Docs were updated.
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/60d338e9-aa44-456a-9ce1-857b7f0d383a)

And I also tried it, and it worked as expected.
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/0e80df94-7f42-4b03-89d0-c225a4a8907f)
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/a42111ad-c9ff-43b1-abab-6aad933e8d8b)

[TC2468 - Farmer Bot Installation with power wake_up_threshold](https://app.testlodge.com/a/26076/projects/40893/suites/221831?expand_section=382401#case_4187430)

PR #819, opened 2024-02-12
update thresholds logic in the farmerbot and create independent checks for each cloud unit utilization

xmonader:
update thresholds logic in the farmerbot and create independent checks for each cloud unit utilization.  

### Description

update thresholds logic in the farmerbot and create independent checks for each cloud unit utilization.  

### Changes

- update thresholds logic in the farmerbot and create independent checks for each cloud unit utilization.  
- helpers to calculate demands
- helpers to select underutilized nodes
- helpers to check if demands is met in the current conditions of used and total resources

### Related Issues

- #804 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### Updated report

Screenshot from 2024-02-26 11-55-20


farmerbot/internal/power.go
```diff
@@ -178,40 +245,76 @@ func (f *FarmerBot) resourceUsageTooLow(sub Substrate, usedResources, totalResou
...
-			node.resources.used.mru + node.resources.used.cru
-		newTotalResources -= node.resources.total.hru + node.resources.total.sru +
-			node.resources.total.mru + node.resources.total.cru
 
-		if newTotalResources == 0 {
-			break
-		}
+		cpNewUsedResources := newUsedResources
+		cpNewTotalResources := newTotalResources
+
```
xmonader, line 268: `nodesLeftOnline` should also be restored on err like cpNewUsedResources

farmerbot/internal/power.go
```diff
@@ -8,6 +8,8 @@ import (
 	"golang.org/x/exp/maps"
 )
 
+var thresholdPercentages = ThresholdPercentages{cru: 200, mru: 80, sru: 80, hru: 80}
```
xmonader: should be moved into configuration

farmerbot/internal/power.go
```diff
@@ -178,40 +245,76 @@ func (f *FarmerBot) resourceUsageTooLow(sub Substrate, usedResources, totalResou
...
-			node.resources.total.mru + node.resources.total.cru
 
-		if newTotalResources == 0 {
-			break
-		}
+		cpNewUsedResources := newUsedResources
+		cpNewTotalResources := newTotalResources
+
+		newUsedResources.cru -= node.resources.used.cru
+		newUsedResources.mru -= node.resources.used.sru
```
rawdaGastan: Mismatched `sru` and `mru`

farmerbot/internal/power.go
```diff
@@ -137,24 +156,78 @@ func calculateResourceUsage(nodes map[uint32]node) (uint64, uint64) {
...
+		return fmt.Errorf("no available nodes to wake up, resources usage is high")
+	}
+
+	for _, node := range nodes {
 		if node.powerState == off {
-			log.Info().Uint32("nodeID", nodeID).Msg("Too much resource usage. Turning on node")
-			return f.powerOn(sub, nodeID)
+			log.Info().Uint32("nodeID", uint32(node.ID)).Msg("Too much resource usage. Turning on node")
+			if err := f.powerOn(sub, uint32(node.ID)); err != nil {
+				return fmt.Errorf("couldn't power on node %v with error: %w", node.ID, err)
```
rawdaGastan: Should we return? Or continue?

xmonader: we should continue indeed you're right

scottyeager:
Anything blocking here? Would be nice to get this merged, since the current implementation means that in many cases new nodes won't be woken when they are needed.

farmerbot/internal/power.go
```diff
@@ -141,23 +168,78 @@ func calculateResourceUsage(nodes map[uint32]node) (uint64, uint64) {
...
-func (f *FarmerBot) resourceUsageTooHigh(sub Substrate) error {
-	for nodeID, node := range f.nodes {
+func (f *FarmerBot) resourceUsageTooHigh(sub Substrate, demand capacity) error {
+	log.Info().Msg("Too high resource usage. Powering on some nodes")
+	nodes, err := f.selectNodesToPowerOn(demand)
+	if err != nil {
+		return err
+	}
+
+	if len(nodes) == 0 {
```
Omarabdul3ziz: selectNodesToPowerOn already return error if can't meet demand, right?

farmerbot/internal/power.go
```diff
@@ -141,23 +168,78 @@ func calculateResourceUsage(nodes map[uint32]node) (uint64, uint64) {
...
+		return fmt.Errorf("no available nodes to wake up, resources usage is high")
+	}
+
+	for _, node := range nodes {
 		if node.powerState == off {
```
Omarabdul3ziz: also it is already skipping the node not off

farmerbot/internal/power.go
```diff
@@ -141,23 +168,78 @@ func calculateResourceUsage(nodes map[uint32]node) (uint64, uint64) {
...
+	var selectedNodes []node
+	remainingDemand := demand
+
+	for _, node := range f.nodes {
+		if node.powerState != off {
+			continue // Skip nodes that are already on or waking up
+		}
+
+		// Check if this node can contribute to the remaining demand
+		contribute := false
```
Omarabdul3ziz, line 184: do you think it is a good idea to find the least number of nodes to wakeup?
i am thinking of in some cases maybe node1 satisfy cru only and it will be added to selectedNodes to power on, then node2 will satisfy the rest of resources and it also can satisfy the initial demand of cru, so instead of powering on both, node2 will be enough.

Omarabdul3ziz (commented):
looks good to me, just some small comments.

Eslam-Nawara (commented):
Nice work, agree with Omar comments :)

farmerbot/internal/power.go
rawdaGastan: I discussed it with @muhamadazmy. I think it is ok (your suggestion is right) to power on the next node that fits requirements. Of course, we don't guarantee if it will be used or not.

Omarabdul3ziz (approved):
LGTM

Issue #828, opened 2024-02-14
Unify workloads metadata across clients


AbdelrahmanElawady:
## Description

As mentioned in https://github.com/threefoldtech/tfgrid-sdk-ts/issues/2164 there are some differences between both clients that cause the user experience to be degraded especially on the dashboard. So, unifying both clients would make it easier for deployments to be fetched and presented across different tools (terraform, cli, dashboard, etc...).

## Suggested changes

The metadata for specific solutions are not handled by go client (Taiga, PeerTube, etc...) as the client treats them all the same as VMs with different flists but it should be the user's responsibility to match the format given in the parent issue in order for it to be listed on the dashboard. But we will handle the default values for different workloads to be listed as just VMs and also to support updating Networks in both clients.

All metadata will also have `version` field, currently set to 3.

### VM

The default metadata for any VM should look like:

```json
{"version": 3, "type":"vm","name":"<name>","projectName":"vm/<name>"}
```

### Kubernetes

The default metadata for any Kubernetes master and workers should look like:

```json
{"version": 3, "type":"kubernetes","name":"<name>","projectName":"<name>"}
```

### Network

Mainly two options:

- Add the IPs used per network workload to its workload metadata.
- Calculate IPs from all deployments for a user on the node that network is on to get this information. This won't require metadata changes but possibly more requests to the node.

Eslam-Nawara:
WIP: 
- update all workloads to have version and set to 3.
- update project name default of vms to be vm/<name>.
- update project name default of k8s to be `.
- update grid-cli and tfrobot to use new project name format for vms.

Eslam-Nawara:
WIP: 
Testing the new update for compatibility with deployments with the old versions

- Creating deployment with the old format:
```console
➜  grid-cli git:(development) ./bin/tfcmd deploy vm --name cliTest --ssh ~/.ssh/id_rsa.pub
11:52AM INF starting peer session=tf-155307 twin=4653
11:52AM INF deploying network
11:53AM INF deploying vm
11:53AM INF vm planetary ip: 302:9e63:7d43:b742:85ef:2cbe:d554:b95b
```

- Loading the deployment with the new version:
```console 
➜  grid-cli git:(development-update-workloads-metadata) ./bin/tfcmd get vm cliTest
11:54AM INF starting peer session=tf-158414 twin=4653
11:54AM INF vm:
{
        "Name": "cliTest",
        "NodeID": 11,
[...]

PR #848, opened 2024-02-20
Update workloads metadata

Eslam-Nawara:
### Description

update workloads meatadata to be Unified across all clients

### Changes

- update all workloads to have version and set to 3.
- update project name default of vms to be `vm/<name>`.
- update project name default of k8s to be `<name>.
- update grid-cli and tfrobot to use new project name format for vms.
- restore tfrobot examples

### Related Issues

- #828 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation


grid-cli/internal/cmd/get.go
```diff
@@ -10,31 +10,43 @@ import (
...
 		return workloads.Deployment{}, err
 	}
 
-	networkContractIDs, err := t.ContractsGetter.GetNodeContractsByTypeAndName(name, workloads.NetworkType, fmt.Sprintf("%snetwork", name))
+	networkContractIDs, err := t.ContractsGetter.GetNodeContractsByTypeAndName(projectName, workloads.NetworkType, fmt.Sprintf("%snetwork", name))
 	if err != nil {
 		return workloads.Deployment{}, err
 	}
 
+	if len(nodeContractIDs) == 0 && len(networkContractIDs) == 0 {
```
AbdelrahmanElawady: This won't happen as `GetNodeContractsByTypeAndName` will fail if there are no contracts which will make it only check new format and fail if contracts are not found

tfrobot/pkg/deployer/load.go
```diff
@@ -247,3 +226,32 @@ func loadDeploymentWithContractID(ctx context.Context, tfPluginClient deployer.T
 
 	return workloads.NewDeploymentFromZosDeployment(dl, nodeID)
 }
+
+func getContractsWithProjectName(ctx context.Context, tfPluginClient deployer.TFPluginClient, nodeGroup string) ([]uint64, error) {
+	// try to load group with the new name format "vm/<group name>"
+	name := fmt.Sprintf("vm/%s", nodeGroup)
+
+	contracts, err := tfPluginClient.ContractsGetter.ListContractsOfProjectName(name, true)
+	if err != nil || len(contracts.NodeContracts) == 0 {
```
AbdelrahmanElawady: if there is an error it should return immediately, right? that's different than finding no contracts.

tfrobot/pkg/deployer/load.go
```diff
@@ -247,3 +226,32 @@ func loadDeploymentWithContractID(ctx context.Context, tfPluginClient deployer.T
...
+		// if load failed try to load group with the old name format "<group name>"
+		contracts, err = tfPluginClient.ContractsGetter.ListContractsOfProjectName(nodeGroup, true)
+		if err != nil {
+			log.Debug().Err(err).Str("node group", nodeGroup).Msg("couldn't list contracts")
+
+			return nil, fmt.Errorf("couldn't list contracts of node group %s: %s", nodeGroup, err.Error())
```
AbdelrahmanElawady: errors should be wrapped with %w

tfrobot/pkg/deployer/load.go
```diff
@@ -247,3 +226,32 @@ func loadDeploymentWithContractID(ctx context.Context, tfPluginClient deployer.T
...
+
+			return nil, fmt.Errorf("couldn't find any contracts of node group %s", nodeGroup)
+		}
+	}
+
+	ContractIDs, err := getContractIDs(contracts.NodeContracts)
+	if err != nil {
+		log.Debug().Err(err).Str("node group", nodeGroup).Msg("couldn't parse contract id")
+
+		return nil, fmt.Errorf("couldn't parse contract id of node group %s: %s", nodeGroup, err.Error())
```
AbdelrahmanElawady: same here, also without the .Error() part

AbdelrahmanElawady (changes requested):
nice work

grid-cli/internal/cmd/get.go
```diff
@@ -24,17 +24,16 @@ func GetVM(ctx context.Context, t deployer.TFPluginClient, name string) (workloa
...
-	}
+	nodeContractIDs, err := t.ContractsGetter.GetNodeContractsByTypeAndName(projectName, workloads.VMType, name)
+	if err == nil {
+		networkContractIDs, err = t.ContractsGetter.GetNodeContractsByTypeAndName(projectName, workloads.NetworkType, fmt.Sprintf("%snetwork", name))
+		if err != nil {
+			return workloads.Deployment{}, err
+		}
 
-	if len(nodeContractIDs) == 0 && len(networkContractIDs) == 0 {
+	} else {
```
AbdelrahmanElawady: this doesn't tell the difference between an error with fetching/parsing or just no contracts found, I think it shouldn't be handled that way as we only should continue if there are no errors but we didn't find a contract. We can do that with either defining an error in the function we call that we can check on it here or stop it from returning errors on 0 contracts. I prefer defining an error as to not change it's functionality.

Eslam-Nawara: I've added a new error type ` ErrorContractNotFound` in case of empty list of contracts and used it instead of just checking if error is not nil

AbdelrahmanElawady (approved):

Issue #823, opened 2024-02-13
🐞 [Bug] Farmerbot: Validation for running multiple command at once 


A-Harby:
### What happened?

There should be validation to prevent using more than one command.
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/a8a887f1-be27-4bc2-8fdf-41e138356ee0)

### which network/s did you face the problem on?

Dev

### Relevant log output

```console
farmerbot run start -c farmerbot_config.yml -e .env -d
```

xmonader:
Addressed in https://github.com/threefoldtech/tfgrid-sdk-go/pull/849

PR #849, opened 2024-02-20
don't allow subcommands to be used with each other in the farmerbot cli

xmonader:
### Description

don't allow subcommands to be used with each other in the farmerbot cli

### Changes

Converts all of subcommands to use RunE, and do a check on the length of the arguments before the execution and show an error message that has the command name and the other command that got passed and inform the user that they can't be used together 
### Related Issues

#823

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/farmerbot/cmd/start.go
+++ b/farmerbot/cmd/start.go
@@ -13,6 +13,10 @@ var startCmd = &cobra.Command{
 	Use:   "start",
 	Short: "start a node in your farm",
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(cmd.Flags().Args()) != 0 {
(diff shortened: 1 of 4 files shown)


farmerbot/cmd/start.go
```diff
@@ -13,6 +13,10 @@ var startCmd = &cobra.Command{
 	Use:   "start",
 	Short: "start a node in your farm",
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(cmd.Flags().Args()) != 0 {
```
rawdaGastan, line 16: `start` has `all` as a sub command

rawdaGastan: It doesn't check the `start` command in case I used `all` so this is right :D

rawdaGastan (approved):

Issue #857, opened 2024-02-26
Refactor `NewTFPluginClient` to use options instead of long list of optional arguments


AbdelrahmanElawady:
## Description 

Refactor `NewTFPluginClient` to use options instead of long list of optional arguments

AbdelrahmanElawady:
Verified:

`NewTFPluginClient` now takes optional arguments:

```go
deployer.NewTFPluginClient(
	mnemonic,
	deployer.WithKeyType(peer.KeyTypeSr25519),
	deployer.WithLogs(),
	deployer.WithNetwork("main"),
	deployer.WithTwinCache(),
)
```

PR #863, opened 2024-02-27
Use options for optional arguments in `NewTFPluginClient`

Eslam-Nawara:
### Description

refactor `Grid Client`

### Changes

- change `NewTFPluginClient` arguments list to use options instead of zero values for optional arguments.
- update the `substrateURL` argument to be list of URLs as multiple substrate URLs are supported.

### Related Issues

- #857 

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


grid-client/deployer/tf_plugin_client.go
```diff
@@ -95,20 +95,81 @@ type TFPluginClient struct {
...
+		o(&cfg)
+	}
+
+	return cfg
+}
+
 // NewTFPluginClient generates a new tf plugin client
 func NewTFPluginClient(
 	mnemonicOrSeed string,
 	keyType string,
```
AbdelrahmanElawady: keyType should be an option too with the default being `Sr25519` like Peer, right?

grid-client/deployer/tf_plugin_client.go
```diff
@@ -95,20 +95,81 @@ type TFPluginClient struct {
...
+	relayURL      string
+	proxyURL      string
+	rmbTimeout    int
+	showLogs      bool
+	rmbInMemCache bool
+}
+
+type PluginOpt func(*pluginCfg)
+
+func WithSubstrateURL(substrateURL []string) PluginOpt {
```
AbdelrahmanElawady:
```suggestion
func WithSubstrateURL(substrateURL ...string) PluginOpt {
```

would that be better so user can pass one or more without wrapping it in a slice.
so both can work: `WithSubstrateURL("url")` and `WithSubstrateURL(someSlice...)` instead of `WithSubstrateURL([]string{"url"})`

AbdelrahmanElawady (commented):
nice work

grid-client/deployer/tf_plugin_client.go
```diff
@@ -95,20 +95,81 @@ type TFPluginClient struct {
...
+		p.rmbInMemCache = true
+	}
+}
+
+func parsePluginOpts(opts ...PluginOpt) pluginCfg {
+	cfg := pluginCfg{
+		substrateURL:  []string{},
+		proxyURL:      "",
+		relayURL:      "",
+		rmbTimeout:    60, // default rmbTimeout is 60
```
rawdaGastan, line 166: why 60? I think 1 min is too long? no?

grid-client/deployer/tf_plugin_client.go
```diff
@@ -95,20 +95,81 @@ type TFPluginClient struct {
 	cancelRelayContext context.CancelFunc
 }
 
+type pluginCfg struct {
```
rawdaGastan, line 98: I think we should add the network and key type too

grid-client/deployer/tf_plugin_client.go
```diff
@@ -95,20 +95,81 @@ type TFPluginClient struct {
...
+}
+
+func WithRMBInMemCache() PluginOpt {
+	return func(p *pluginCfg) {
+		p.rmbInMemCache = true
+	}
+}
+
+func parsePluginOpts(opts ...PluginOpt) pluginCfg {
+	cfg := pluginCfg{
```
rawdaGastan, line 160: add the default values for all

grid-client/deployer/tf_plugin_client.go
```diff
@@ -152,11 +213,13 @@ func NewTFPluginClient(
 	tfPluginClient.Network = network
 
 	tfPluginClient.substrateURL = SubstrateURLs[network]
-	if len(strings.TrimSpace(substrateURL)) != 0 {
-		if err := validateWssURL(substrateURL); err != nil {
-			return TFPluginClient{}, errors.Wrapf(err, "could not validate substrate url %s", substrateURL)
+	if len(cfg.substrateURL) != 0 {
```
rawdaGastan: we can add the validations in `with..` functions

grid-client/deployer/tf_plugin_client.go
```diff
@@ -95,20 +95,81 @@ type TFPluginClient struct {
...
+}
+
+func parsePluginOpts(opts ...PluginOpt) pluginCfg {
+	cfg := pluginCfg{
+		substrateURL:  []string{},
+		proxyURL:      "",
+		relayURL:      "",
+		rmbTimeout:    60, // default rmbTimeout is 60
+		showLogs:      false,
+		rmbInMemCache: false,
```
rawdaGastan: mmm do you think we should set the default with `true`?

grid-cli/cmd/cancel.go
```diff
@@ -20,7 +20,12 @@ var cancelCmd = &cobra.Command{
 		if err != nil {
 			log.Fatal().Err(err).Send()
 		}
-		t, err := deployer.NewTFPluginClient(cfg.Mnemonics, "sr25519", cfg.Network, "", "", "", 100, false, true)
+
+		opts := []deployer.PluginOpt{
+			deployer.WithRMBInMemCache(),
```
rawdaGastan: we can put them in the `NewTFPluginClient` function itself. check all

grid-client/deployer/network_deployer_test.go
```diff
@@ -164,7 +164,11 @@ func ExampleNetworkDeployer_Deploy() {
 	const network = "<dev, test, qa, main>"
 	const nodeID = 11 // use any node with status up, use ExampleFilterNodes to get valid nodeID
 
-	tfPluginClient, err := NewTFPluginClient(mnemonic, "sr25519", network, "", "", "", 0, false, true)
+	opts := []PluginOpt{
+		WithRMBInMemCache(),
+		WithRMBTimeout(10),
```
rawdaGastan: why?

rawdaGastan (changes requested):

grid-client/deployer/tf_plugin_client.go
```diff
@@ -95,27 +95,126 @@ type TFPluginClient struct {
...
+	for _, o := range opts {
+		o(&cfg)
+	}
+
+	if strings.TrimSpace(cfg.proxyURL) == "" {
+		cfg.proxyURL = ProxyURLs[cfg.network]
+
+		if err := validateProxyURL(cfg.proxyURL); err != nil {
+			return cfg, errors.Wrapf(err, "could not validate proxy url %s", cfg.proxyURL)
+		}
```
AbdelrahmanElawady: this validation will only happen if the user didn't provide a url, right? we want to validate the user input too

grid-client/deployer/tf_plugin_client.go
```diff
@@ -95,27 +95,126 @@ type TFPluginClient struct {
...
+	}
+
+	if strings.TrimSpace(cfg.relayURL) == "" {
+		cfg.relayURL = RelayURLS[cfg.network]
+	}
```
AbdelrahmanElawady, line 184: should be validated that it's a wss url too, right?

grid-client/deployer/tf_plugin_client.go
```diff
@@ -95,27 +95,126 @@ type TFPluginClient struct {
...
+
+	if len(cfg.substrateURL) == 0 {
+		cfg.substrateURL = SubstrateURLs[cfg.network]
+
+		for _, url := range cfg.substrateURL {
+			if err := validateWssURL(url); err != nil {
+				return cfg, errors.Wrapf(err, "could not validate substrate url %s", url)
+			}
+		}
```
AbdelrahmanElawady: same thing here validation only happens when user don't use the option

AbdelrahmanElawady (changes requested):

AbdelrahmanElawady (approved):

rawdaGastan (approved):

Issue #858, opened 2024-02-26
Refactor Network state


AbdelrahmanElawady:
### Is your feature request related to a problem? Please describe

Network state is not needed anymore since we added list_private_ips call: https://github.com/threefoldtech/zos/pull/2206

### Describe the solution you'd like

Refactor local state to not use Network state anymore.

### Related issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/869
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/868
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/904

PR #866, opened 2024-02-28
refactor network state grid client

rawdaGastan:
### Changes

- Replacing host IDs in a local state with zos call for listing private ips
- refactor generate versionless deployment to match the new function of assigning new ips
- refactor assign vms ips 
- update tests

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/858
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/868

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

### TODO

- refactor k8s after the changes (assign ips)
- refactor subnets of network state (needs refactoring the network deployer)


grid-client/deployer/deployment_deployer.go
```diff
@@ -29,102 +30,137 @@ func NewDeploymentDeployer(tfPluginClient *TFPluginClient) DeploymentDeployer {
...
+				return
+			}
+
+			mu.Lock()
+			defer mu.Unlock()
+			gridDlsPerNodes[dl.NodeID] = append(gridDlsPerNodes[dl.NodeID], newDl)
+		}(dl)
+	}
+
+	wg.Wait()
```
AbdelrahmanElawady: I don't think any of these are blocking, do we need concurrency for it?

grid-client/deployer/deployment_deployer.go
```diff
@@ -29,102 +30,137 @@ func NewDeploymentDeployer(tfPluginClient *TFPluginClient) DeploymentDeployer {
...
-			if _, ok := newDeployments[nodeID]; !ok {
-				newDeploymentsSolutionProvider[nodeID] = []*uint64{dl.SolutionProvider}
-				newDeployments[nodeID] = []gridtypes.Deployment{generatedDl}
-				continue
-			}
-			newDeployments[nodeID] = append(newDeployments[nodeID], generatedDl)
-			newDeploymentsSolutionProvider[nodeID] = append(newDeploymentsSolutionProvider[nodeID], dl.SolutionProvider)
-		}
+	if err := d.Validate(ctx, dls); err != nil {
+		return fmt.Errorf("invalid deployment: %w", err)
```
AbdelrahmanElawady: I think we want `BatchDeploy` to be non-atomic, so this will fail on one invalid deployment, right?

grid-client/deployer/deployment_deployer.go
```diff
@@ -29,102 +30,137 @@ func NewDeploymentDeployer(tfPluginClient *TFPluginClient) DeploymentDeployer {
...
+	}
+
+	newDeployments, err := d.GenerateVersionlessDeployments(ctx, dls)
+	if err != nil {
+		return fmt.Errorf("could not generate grid deployments: %w", err)
```
AbdelrahmanElawady: same here

grid-client/deployer/deployment_deployer.go
```diff
@@ -185,24 +208,189 @@ func (d *DeploymentDeployer) updateStateFromDeployments(ctx context.Context, dl
...
+
+	// calculate used host IDs per network
+	for _, dl := range dls {
+		mu.Lock()
+		if _, ok := usedHosts[dl.NetworkName]; !ok {
+			usedHosts[dl.NetworkName] = make(map[uint32][]byte)
+		}
+
+		if _, ok := usedHosts[dl.NetworkName][dl.NodeID]; ok {
+			continue
```
AbdelrahmanElawady, line 278: if we continue here with the mutex still locked it can be a deadlock as it will never be freed

grid-client/deployer/deployment_deployer.go
```diff
@@ -185,24 +208,189 @@ func (d *DeploymentDeployer) updateStateFromDeployments(ctx context.Context, dl
...
+				defer mu.Unlock()
+				errs = multierror.Append(errs, errors.Wrapf(err, "invalid ip range %s", ipRange))
+				return
+			}
+
+			// get used host IDs of the deployment node using the private IPs list of the node
+			for _, privateIP := range privateIPs {
+				parsedPrivateIP := net.ParseIP(privateIP).To4()
+
+				if ipRangeCIDR.Contains(parsedPrivateIP) {
```
AbdelrahmanElawady: if that didn't happen it should be an error, right? a private IP on a node should always be on the node subnet. maybe we should log it if it happened?

grid-client/deployer/deployment_deployer.go
```diff
@@ -185,24 +208,189 @@ func (d *DeploymentDeployer) updateStateFromDeployments(ctx context.Context, dl
...
+					mu.Lock()
+					defer mu.Unlock()
+					usedHosts[dl.NetworkName][dl.NodeID] = append(usedHosts[dl.NetworkName][dl.NodeID], parsedPrivateIP[3])
```
AbdelrahmanElawady: we can move all that to a new function that return error and used hosts so we can lock once instead of locking appending unlocking returning multiple times

grid-client/deployer/deployment_deployer.go
```diff
@@ -185,24 +208,189 @@ func (d *DeploymentDeployer) updateStateFromDeployments(ctx context.Context, dl
...
+			if err != nil {
+				mu.Lock()
+				defer mu.Unlock()
+				errs = multierror.Append(errs, errors.Wrapf(err, "invalid ip range %s", ipRange))
+				return
+			}
+
+			// get used host IDs of the deployment node using the private IPs list of the node
+			for _, privateIP := range privateIPs {
+				parsedPrivateIP := net.ParseIP(privateIP).To4()
```
AbdelrahmanElawady: To4() returns nil if it is not a valid ipv4, maybe we should check that before accessing the ip so it won't panic

grid-client/deployer/deployment_deployer.go
```diff
@@ -185,24 +208,189 @@ func (d *DeploymentDeployer) updateStateFromDeployments(ctx context.Context, dl
...
+	for _, dl := range dls {
+		if len(dl.Vms) == 0 {
+			continue
 		}
-		network := d.tfPluginClient.State.Networks.GetNetwork(dl.NetworkName)
-		updateNetworkUsedIPs(&network, dl)
-		d.tfPluginClient.State.Networks[dl.NetworkName] = network
+
+		wg.Add(1)
+		go func(dl *workloads.Deployment) {
```
AbdelrahmanElawady, line 319: here also there is no blocking calls. I don't thing goroutines would perform better

AbdelrahmanElawady (changes requested):
Amazing work! I think this should go to `development_3.14` as this is still didn't reach mainnet

grid-client/deployer/deployment_deployer.go
```diff
@@ -29,102 +30,141 @@ func NewDeploymentDeployer(tfPluginClient *TFPluginClient) DeploymentDeployer {
...
+			defer mu.Unlock()
+			gridDlsPerNodes[dl.NodeID] = append(gridDlsPerNodes[dl.NodeID], newDl)
+		}(dl)
 	}
 
-	return map[uint32]gridtypes.Deployment{dl.NodeID: newDl}, usedHosts, nil
+	wg.Wait()
+
+	if errs != nil {
+		return nil, errs
```
AbdelrahmanElawady: I don't think we should return empty map of deployments. There might be some deployments that managed to be generated. We should return everything and the caller decide if they want to continue with the deployment

grid-client/deployer/deployment_deployer.go
```diff
@@ -29,102 +30,141 @@ func NewDeploymentDeployer(tfPluginClient *TFPluginClient) DeploymentDeployer {
...
-	network := d.tfPluginClient.State.Networks.GetNetwork(dl.NetworkName)
-	newDeployments, _, err := d.GenerateVersionlessDeployments(ctx, dl, network.GetUsedNetworkHostIDs(dl.NodeID))
+	dlsPerNodes, err := d.GenerateVersionlessDeployments(ctx, []*workloads.Deployment{dl})
 	if err != nil {
 		return errors.Wrap(err, "could not generate deployments data")
 	}
 
-	dl.NodeDeploymentID, err = d.deployer.Deploy(ctx, dl.NodeDeploymentID, newDeployments, newDeploymentsSolutionProvider)
+	if len(dlsPerNodes[dl.NodeID]) == 0 {
+		return errors.Wrap(err, "failed to generate the grid deployment")
```
AbdelrahmanElawady: this will actually return nil, as `errors.Wrap` returns a nil error if `err` is nil

grid-client/deployer/deployment_deployer.go
```diff
@@ -29,102 +30,141 @@ func NewDeploymentDeployer(tfPluginClient *TFPluginClient) DeploymentDeployer {
...
+		multiErr = multierror.Append(multiErr, fmt.Errorf("invalid deployments: %w", err))
+	}
+
+	newDeployments, err := d.GenerateVersionlessDeployments(ctx, dls)
+	if err != nil {
+		multiErr = multierror.Append(multiErr, fmt.Errorf("could not generate grid deployments: %w", err))
+	}
+
+	if len(newDeployments) == 0 {
+		return errors.Wrap(err, "failed to generate the grid deployments")
```
AbdelrahmanElawady: same here, also we should use multiErr not err

grid-client/deployer/deployment_deployer.go
```diff
@@ -185,24 +212,204 @@ func (d *DeploymentDeployer) updateStateFromDeployments(ctx context.Context, dl
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "invalid ip range %s", ipRange)
+	}
+
+	// get used host IDs of the deployment node using the private IPs list of the node
+	for _, privateIP := range privateIPs {
+		parsedPrivateIP := net.ParseIP(privateIP).To4()
+
+		if parsedPrivateIP == nil {
+			log.Debug().Err(fmt.Errorf("[Error]: network %s has a nil private ip %s", networkName, privateIP)).Send()
```
AbdelrahmanElawady: it's a nitpick but it's not necessarily a nil private ip but could be any random characters that are not valid ipv4 :D

AbdelrahmanElawady (changes requested):

grid-client/deployer/deployment_deployer.go
rawdaGastan: why should I continue with no deployments?

AbdelrahmanElawady: no no you shouldn't continue, I meant if err here is nil it will also return nil, and multiErr should have both errors from validation and from generating deployments so it would be better to use it.

AbdelrahmanElawady (approved):

Issue #911, opened 2024-03-05
Support more than one filter together in proxy.


AlaaElattar:
### Is your feature request related to a problem? Please describe

- For example i wanna get all my rent contracts, but i wanna show both `Created` and `GracePeriod`.
- but how to do that and paginate them if i only get one of them in the request ?

### Describe the solution you'd like

- I'd like to filter and get both my contracts `Created` and `GracePeriod` together in one call.

Eslam-Nawara:
Verified:
requested contracts in created and grace period states and the filtration worked as expected, the list had contracts with both Created and in GracePeriod states and no contracts in Deleted state
https://gridproxy.dev.grid.tf/contracts?size=9999&state=GracePeriod,Created 

![Screenshot from 2024-06-10 16-07-11](https://github.com/threefoldtech/tfgrid-sdk-go/assets/67752395/2ad71036-8751-4ae6-8914-f1347643dbcf)

PR #920, opened 2024-03-06
allow filtering with array of contract states

Omarabdul3ziz:
### Description
allow filtering with array of contract states

### Changes
- register a custome Encoder/Decoder for in the schema pkg for []string param
- change the type of ContractFilter to be []string
- modify the sql query/ mock tests to check the existing of the state in an array instead of exact equal

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/911

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation


grid-proxy/pkg/client/grid_client.go
```diff
@@ -8,15 +8,29 @@ import (
...
+var encoder *schema.Encoder
+
+func init() {
+	encoder = schema.NewEncoder()
+
+	encoder.RegisterEncoder([]string{}, func(value reflect.Value) string {
+		if value.Kind() == reflect.Slice && value.Type().Elem().Kind() == reflect.String {
+			slice := value.Interface().([]string)
+			return strings.Join(slice, ",")
+		}
```
AbdelrahmanElawady: I am curious why we are checking value type and its elements type, doesn't the first value in `RegisterEncoder` guarantee only []string will be passed?
Also, wouldn't something like `slice, ok := slice := value.Interface().([]string)` work instead of the multiple reflect calls?

grid-proxy/tests/queries/mock_client/contracts.go
```diff
@@ -96,7 +97,7 @@ func (c *RentContract) satisfies(f types.ContractFilter) bool {
 		return false
 	}
 
-	if f.State != nil && *f.State != c.State {
+	if f.State != nil && len(f.State) != 0 && !containsState(f.State, c.State) {
```
AbdelrahmanElawady: I think neither `f.State != nil` or `len(f.State) != 0` is needed, `containsState` should be enough

grid-proxy/internal/explorer/db/postgres.go
```diff
@@ -783,8 +783,13 @@ func (d *PostgresDatabase) GetContracts(ctx context.Context, filter types.Contra
 	if filter.Type != nil {
 		q = q.Where("type = ?", *filter.Type)
 	}
-	if filter.State != nil {
-		q = q.Where("state ILIKE ?", *filter.State)
+	if filter.State != nil && len(filter.State) != 0 {
```
AbdelrahmanElawady: no need to check for nil `len` works fine with nil

AbdelrahmanElawady (commented):
Looks good, just some questions

AbdelrahmanElawady (approved):

Issue #926, opened 2024-03-07
query/filter nodes node has_ipv6 


Omarabdul3ziz:
### Is your feature request related to a problem? Please describe

query/filter nodes node has_ipv6 

### Describe the solution you'd like

data will be indexed directly from the node with the rmb command `zos.network.has_ipv6`

AhmedHanafy725:
please also add a field for it on the nodes' response

PR #937, opened 2024-03-11
add new indexer work to check node having ipv6

Omarabdul3ziz:
### Description
add new indexer work to check node having ipv6

### Changes
- add generator/loader for twin ipv6 and update the schema for the new table `node_ipv6`
- add handle tests for the new filter on nodes `has_ipv6`
- add new indexer work to call up nodes and check if they have ipv6
- update the db filters with the `has_ipv6` filter on nodes
- add default values for the flags (1 worker runs every 1 day)
- update makefile with helpful command to get db dump from the dev cluster

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/926

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation


grid-proxy/Makefile
```diff
@@ -3,6 +3,11 @@ PQ_HOST = $(shell docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddr
 PQ_CONTAINER = postgres
 count = 3
 
+NAMESPACE := tfchain-dev
+DB_POD := $(shell kubectl --namespace $(NAMESPACE) get pods | grep processor-db | awk '{print $$1}')
+DUMP_CMD := PGPASSWORD="postgres" pg_dump -U postgres tfgrid-graphql > /tmp/dump.sql
```
xmonader: اتقوا الله

grid-proxy/tools/db/crafter/generator.go
```diff
@@ -937,3 +937,30 @@ func (c *Crafter) GeneratePricingPolicies() error {
 
 	return err
 }
+
+func (c *Crafter) GenerateNodeIpv6() error {
```
AbdelrahmanElawady, line 941: I don't see this being used in `generateData`. It should be, right?

sameh-farouk:
Thank you!
I just want to ask why not do this in the GraphQL processor and have real-time data?
We can save this data in a separate table when a node is created or updated.
Also, instead of creating indexes on the indexer, we can define them in the GraphQL schema. This is as simple as adding a few characters to the graphql schema file.
In my opinion, it may be more effective to maximize the utilization of the current solution rather than continuing to implement additional layers of complexity to accomplish similar tasks.
If a task can be performed by the processor, it is completed there. Otherwise, we move it to another layer.

Omarabdul3ziz:
afaik the chain doesn't have this info, does it? and we need to ask the node for it.
so we will need to add a service on the graphql that asks the node for this info, which we already have on the proxy.

sameh-farouk:
> afaik the chain doesn't have this info, does it? and we need to ask the node for it. so we will need to add a service on the graphql that asks the node for this info, which we already have on the proxy.

Is it something different than the ips in the interface in tfgridmodule ?
```
{
  version: 5
  id: 29
  farmId: 180
  twinId: 515
  resources: {
    hru: 2,000,398,934,016
    sru: 128,035,676,160
    cru: 4
    mru: 16,693,874,688
  }
  location: {
    city: Unknown
    country: Belgium
    latitude: 50.8509
    longitude: 4.3447
  }
  publicConfig: null
  created: 1,709,817,558
  farmingPolicyId: 1
  interfaces: [
    {
      name: zos
      mac: 00:e0:4c:68:0a:82
      ips: [
        10.4.0.77
        2a02:1802:5e:14:4915:d057:abdd:322f
      ]
    }
  ]
  certification: Diy
  secureBoot: false
  virtualized: false
  serialNumber: Default string
  connectionPrice: 80
}
```

AbdelrahmanElawady (approved):

Omarabdul3ziz:
i thought it was different, but checking the data on chain for different node and the responses of the rmb-call `zos.network.has_ipv6` looks it do report for the existence of the IP on the chain response. but i am not sure, @muhamadazmy is this true?

Omarabdul3ziz:
checked with azmy and it turns out they are different:
the call `zos.network.has_ipv6` checks if the node can provide ipv6 for the vms running on it.
but the ips on the interfaces are for the node itself.

Issue #917, opened 2024-03-06
add support for multiple rmb relays in the rmb sdk


muhamadazmy:
### Is your feature request related to a problem? Please describe

rmb-peer supports connecting to multiple relays for redundancy. This feature allows the rmb-peer to continue operating if one of the relays is facing technical problems.

It also allows messages to be balanced between multiple relays.

### Describe the solution you'd like

Support for multiple relays need to also be implemented in the go client

rawdaGastan:
### WIP

- Investigating twin updates on different relays when multiple relays are supported

Omarabdul3ziz:
verified:
used my mnemonic which has only one relay configured to make an rmb call using the rmb-sdk peer and added two relays with the option `WithRelay()`  
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/32527347/d1087549-3a34-40d1-883e-b77c1de950b2)

and my twin relays is updated, the two relays hostnames separated by `_`
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/32527347/65d7ec9c-a9fb-4d77-b46c-d8924b2b2d08)

PR #947, opened 2024-03-12
support multiple relays in rmb client

rawdaGastan:
### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/917


rmb-sdk-go/peer/connection.go
```diff
@@ -157,23 +161,49 @@ func (c *InnerConnection) connect() (*websocket.Conn, error) {
...
-	if resp.StatusCode != http.StatusSwitchingProtocols {
-		return nil, fmt.Errorf("invalid response %s", resp.Status)
-	}
+		return nil
+	}, b, func(err error, _ time.Duration) {
+		log.Error().Err(err).Msg("failed to connect to relay, retrying")
+	})
 
-	return con, nil
+	return con, err
```
sameh-farouk: This will open a connection to a single relay and return with the first successful attempt. This approach will create issues when delivering messages to the client. 
Imagine having two email addresses that your friends can use any of them to message you, but you only check one of them.

Instead, you should enable the client to open multiple connections to relays simultaneously, And constantly listen for anything passed over any of these connections. If a connection to one of the relays fails to establish, the client should asynchronously keep trying as long as it is running. And if a connection to one of the relays drops, the client should keep attempt to reconnect.

sameh-farouk (changes requested):

rmb-sdk-go/peer/connection.go
```diff
@@ -146,37 +148,57 @@ func (c *InnerConnection) Start(ctx context.Context) (Reader, Writer) {
...
 
-	return c.loop(ctx, con, output, input)
+	var m sync.Mutex
+	var wg sync.WaitGroup
+
+	for _, con := range connections {
+		wg.Add(1)
+		go func(con *websocket.Conn) {
+			defer wg.Done()
+			m.Lock()
```
AbdelrahmanElawady: this will lead to the first connection acquiring the lock and since `c.loop` is a blocking call it won't release the lock and other connections won't be able to start their connection loop. which means it's still a single relay connection used and will switch when an error happen (the lock released).
we can move `c.loop` to be before the lock or imo using channels will coordinate the different goroutines better.

AbdelrahmanElawady (changes requested):

grid-client/deployer/tf_plugin_client.go
```diff
@@ -47,11 +47,11 @@ var (
 		"main": "https://graphql.grid.tf/graphql",
 	}
 	// RelayURLS relay urls
-	RelayURLS = map[string]string{
-		"dev":  "wss://relay.dev.grid.tf",
-		"test": "wss://relay.test.grid.tf",
-		"qa":   "wss://relay.qa.grid.tf",
-		"main": "wss://relay.grid.tf",
+	RelayURLS = map[string][]string{
```
muhamadazmy: Use the same ones defined here 
https://github.com/threefoldtech/zos/blob/main/pkg/environment/environment.go#L103

rmb-sdk-go/peer/connection.go
```diff
@@ -73,7 +81,28 @@ func (c *InnerConnection) reader(ctx context.Context, cancel context.CancelFunc,
 	}
 }
 
-func (c *InnerConnection) loop(ctx context.Context, con *websocket.Conn, output, input chan []byte) error {
+func (c *InnerConnection) send(ctx context.Context, data []byte) error {
+	if c.busy {
```
muhamadazmy: I suggest u use an atomic bool like here https://pkg.go.dev/sync/atomic#Bool so we don't run into race conditions

rmb-sdk-go/peer/connection.go
```diff
@@ -103,9 +132,26 @@ func (c *InnerConnection) loop(ctx context.Context, con *websocket.Conn, output,
...
 			lastPong = time.Now()
-		case data := <-input:
-			if err := con.WriteMessage(websocket.BinaryMessage, data); err != nil {
-				return err
+		case sent := <-c.writer:
+			// should we remove this flag?
+			c.busy = true
+			defer func() {
+				c.busy = false
+			}()
```
muhamadazmy: You can't use a `defer` in a loop. this is wrong. Read about it

This is why u still had to set c.busy=false later in the code at line (152)

I suggest that u move this entire branch logic to a function so u can use the defer to release the busy flag.

muhamadazmy: As i remember the busy flag was intended to make a quick return if the connection is not established so senders can already try another connection until this one is restored. It means it shouldn't be turned on/off inside the <Send> branch, but instead ONLY when that connection instance is reconnecting to relay.

rmb-sdk-go/peer/peer.go
```diff
@@ -462,19 +483,25 @@ func (d *Peer) makeEnvelope(id string, dest uint32, session *string, cmd *string
...
 	if err != nil {
 		return err
 	}
 
-	select {
-	case d.writer <- bytes:
-	case <-ctx.Done():
-		return ctx.Err()
+	var errs error
+	for _, con := range d.cons {
```
muhamadazmy: may be should be a ring instead of a list, so on next send u try to use another relay? to balance message sending over multiple relays! Not sure if this is the best approach

muhamadazmy (changes requested):

rmb-sdk-go/peer/peer.go
rawdaGastan: Result of connections used after using ring: 

```
con.url: wss://relay.dev.grid.tf
con.url: wss://relay.02.dev.grid.tf
con.url: wss://relay.dev.grid.tf
con.url: wss://relay.02.dev.grid.tf
con.url: wss://relay.dev.grid.tf
con.url: wss://relay.02.dev.grid.tf
con.url: wss://relay.dev.grid.tf
con.url: wss://relay.02.dev.grid.tf
con.url: wss://relay.dev.grid.tf
con.url: wss://relay.02.dev.grid.tf
```

rawdaGastan:
Note: for the atomic boolean before and after connect, connection tries to send before connection so, It always returns it is busy 
I left it in the code to be discussed

muhamadazmy (approved):

PR #951, opened 2024-03-14
fix grid-client network state

Eslam-Nawara:
### Description

- tried to test the client update and found multiple races while deployment

### Changes

- added a lock to network state to prevent any races to the state map
- update `DeploymentDeployer.BatchDeploy` to append errors and not return

### Related Issues

List of related issues

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/grid-client/deployer/deployment_deployer.go
+++ b/grid-client/deployer/deployment_deployer.go
@@ -309,7 +309,7 @@ func (d *DeploymentDeployer) assignPrivateIPs(ctx context.Context, dls []*worklo
 
 	usedHosts, err := d.calculateNetworksUsedIPs(ctx, dls)
 	if err != nil {
-		return errors.Wrap(err, "couldn't calculate network used ips")
+		errs = multierror.Append(errs, errors.Wrap(err, "couldn't calculate networks used ips"))
 	}
 
 	for _, dl := range dls {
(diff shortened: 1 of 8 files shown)


grid-client/state/network_state.go
```diff
@@ -1,10 +1,17 @@
...
 
-// NetworkState is a map of of names and their networks
-type NetworkState map[string]Network
+	"github.com/threefoldtech/zos/pkg/gridtypes"
+)
+
+// NetworkState is a struct of networks names and their networks and mutex to protect the state
+type NetworkState struct {
+	State     map[string]Network
+	StateLock sync.Mutex
```
rawdaGastan: Should be private to prevent outer locks

rawdaGastan (changes requested):

grid-client/deployer/deployment_deployer.go
```diff
@@ -390,11 +391,15 @@ func (d *DeploymentDeployer) assignPrivateIPs(ctx context.Context, dls []*worklo
 			}
 
 			dl.IPrange = ipRange
+
+			mu.Lock()
+			newdls = append(newdls, dl)
```
rawdaGastan, line 397: I think we should append in line 317 too (no vms case)

rawdaGastan (approved):

Issue #918, opened 2024-03-06
Allow sorting nodes with less provisioned cpu


AhmedHanafy725:
### Is your feature request related to a problem? Please describe

 To offer nodes with less CPU usage to the user on the dashboard for the deployments.

### Describe the solution you'd like

 maybe as @Omarabdul3ziz suggested with used < total cpus

Eslam-Nawara:
Verified:
used `sdk-go` to request nodes with limit filter `"sort_by"` = `"free_cru"` and calculated `free_cru` and it's sorted as requested

- returned nodes 
 ```nodeID: 28, free cpu: -51
nodeID: 11, free cpu: -33
nodeID: 177, free cpu: -8
nodeID: 174, free cpu: -8
nodeID: 150, free cpu: -8
nodeID: 128, free cpu: -7
nodeID: 171, free cpu: -5
nodeID: 14, free cpu: -5
nodeID: 163, free cpu: -5
nodeID: 127, free cpu: -4
```

- without using the sort_by filter
```nodeID: 11, free cpu: -33
nodeID: 14, free cpu: -5
nodeID: 28, free cpu: -51
nodeID: 127, free cpu: -4
nodeID: 128, free cpu: -7
nodeID: 144, free cpu: -1
nodeID: 145, free cpu: -3
nodeID: 146, free cpu: -1
nodeID: 150, free cpu: -8
nodeID: 152, free cpu: -2
```

PR #961, opened 2024-03-18
allow sorting nodes with the free_cru

Omarabdul3ziz:
### Description
allow sorting nodes with the free_cru

### Changes
- add new sort param `free_cru`
- add new order query to sort based on `total_cru-used_cru`

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/918

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/grid-proxy/pkg/types/nodes.go
+++ b/grid-proxy/pkg/types/nodes.go
@@ -48,6 +48,7 @@ type Node struct {
 	ExtraFee          uint64       `json:"extraFee" sort:"extra_fee"`
 	Healthy           bool         `json:"healthy"`
 	PriceUsd          float64      `json:"price_usd" sort:"price_usd"`
+	_                 string       `sort:"free_cru"`
 }
 
 // CapacityResult is the NodeData capacity results to unmarshal json in it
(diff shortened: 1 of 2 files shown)


grid-proxy/pkg/types/nodes.go
```diff
@@ -48,6 +48,7 @@ type Node struct {
 	ExtraFee          uint64       `json:"extraFee" sort:"extra_fee"`
 	Healthy           bool         `json:"healthy"`
 	PriceUsd          float64      `json:"price_usd" sort:"price_usd"`
+	_                 string       `sort:"free_cru"`
```
AbdelrahmanElawady, line 51: I am curious as what that does

AbdelrahmanElawady (approved):
looks good just one question

grid-proxy/pkg/types/nodes.go
xmonader: What does it do? :D

Omarabdul3ziz: - the validation on the `sort_by` params is done based on the fields that have a `sort` tag on the `Node` struct.
- but adding a `FreeCru` field to the Node struct will mostly show negative values due to the over-provisioning which may be misleading 

 doing it this way will provide `free_cru` as a valid sorting param without `FreeCru` as a Node field

Issue #891, opened 2024-03-03
update health indexer to use zos.system.diagnostics


Omarabdul3ziz:
### Is your feature request related to a problem? Please describe

currently, we decide if the node is healthy or not based on just its response of the rmb call `zos.system.version`

### Describe the solution you'd like

we can use the call `zos.system.diagnostics` instead, and make the node health depend on the `SystemOk` in the response

rawdaGastan:
### Verification

- `zos.system.version` isn't used now and `zos.system.diagnostics` is used instead

PR #968, opened 2024-03-27
change the call for health indexer to use diagnostics instead

Omarabdul3ziz:
### Description
change the call for health indexer to use diagnostics instead

### Changes
- use diagnostics call in health indexer instead of version call
- decide the node healthyness based on healthy field from diagnostics  response
- update zos version to get the new diagnostics type fields

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/891

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


Omarabdul3ziz:
needs this https://github.com/threefoldtech/zos/pull/2247

Omarabdul3ziz:
it is merged now and safe to update on dev

grid-proxy/internal/indexer/health.go
```diff
@@ -31,28 +32,25 @@ func (w *HealthWork) Finders() map[string]time.Duration {
 }
 
 func (w *HealthWork) Get(ctx context.Context, rmb *peer.RpcClient, twinId uint32) ([]types.HealthReport, error) {
-	var response types.HealthReport
-	err := callNode(ctx, rmb, healthCallCmd, nil, twinId, &response)
+	var diagnostics diagnostics.Diagnostics
+	err := callNode(ctx, rmb, healthCallCmd, nil, twinId, &diagnostics)
+	if err != nil {
+		return []types.HealthReport{}, err
```
AbdelrahmanElawady: there are nodes that don't have this call due to update issues in zos, since they don't have this method they will not respond to the rmb call and it will timeout. They should be reported as unhealthy. Will this error return register them as unhealthy?

Omarabdul3ziz: yeah, a return will just print the error and continue, so if it was healthy before it will keep the state.
updated to not return on error and instead set healthy to false

AbdelrahmanElawady (approved):

Omarabdul3ziz:
waiting this change https://github.com/threefoldtech/zos/pull/2247 to reach all nets 

it depends on the `healthy` field in the `zos.system.diagnositcs` call, it is added in `3.11.2` zos release

Issue #978, opened 2024-04-09
Node wake priority


scottyeager:
### Is your feature request related to a problem? Please describe

Currently the farmer has no control over the order in which farmerbot wakes their nodes.

### Describe the solution you'd like

Allow priority for which nodes to wake first to be specified in the config file. This could be as simple as waking them in the order they are written in the file.

A-Harby:
Verified, Qanet and version v0.15.9.

I tested on farm 180 with 3 nodes: 26, 29, and 28, and set the priority first to node 26 and filled node 28 with deployment to pass the threshold.
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/5b2a2ce3-f288-4b68-ac8b-d7d7fb902b00)

And node 26 was the node to wake up after, as I specified in the config file.
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/44a9ecfb-44fc-45eb-8102-98324d3a6a6d)


Then I reset everything and started all over again, but this time with node 29 as the priority node, I did the same, and as expected, the node I specified was woken up.
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/109473873/5bfea2e7-b986-4795-a712-f8d9c044aac7)

[TC2718 - Farmer Bot Installation with priority_nodes](https://app.testlodge.com/a/26076/projects/40893/suites/221831?expand_section=382401#case_4295012)

PR #988, opened 2024-04-21
 prio nodes farmerbot

rawdaGastan:
### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/978


rawdaGastan:
Should we use a slice of nodes instead of a map?

farmerbot/README.md
```diff
@@ -99,6 +99,8 @@ included_nodes: [optional, if no nodes are added then the farmerbot will include
   - "<your node ID to be included, required at least 2>"
 excluded_nodes:
   - "<your node ID to be excluded, optional>"
+priority:
```
Eslam-Nawara: It should be `priority_nodes`

farmerbot/internal/power.go
```diff
@@ -148,7 +148,15 @@ func calculateResourceUsage(nodes map[uint32]node) (uint64, uint64) {
...
-	for nodeID, node := range f.nodes {
+	nodesKeys := make([]uint32, 0, len(f.nodes))
+	for k := range f.nodes {
+		nodesKeys = append(nodesKeys, k)
+	}
+
+	for i := 0; i < len(f.nodes); i++ {
+		nodeID := nodesKeys[i]
+		node := f.nodes[nodeID]
+
```
Eslam-Nawara: Is there a reason this is done this way? I mean, I think this does the same job as before

Eslam-Nawara:
> Should we use a slice of nodes instead of a map?

I think yes, looping over map, saving the keys in array then depending on the order of the keys this won't guarantee the order of the keys to be as expected either

farmerbot/internal/state.go
```diff
@@ -212,55 +212,68 @@ func getNode(
...
 }
 
 // UpdateNode updates a node in the config
 func (s *state) updateNode(node node) error {
-	s.m.Lock()
-	defer s.m.Unlock()
-
-	nodeID := uint32(node.ID)
+	// s.m.Lock()
+	// defer s.m.Unlock()
```
Eslam-Nawara: I think the lock should be added as we're updating the list

farmerbot/internal/farmerbot.go
```diff
@@ -328,11 +333,39 @@ func (f *FarmerBot) iterateOnNodes(ctx context.Context, subConn Substrate) error
 	return nil
 }
 
+func addPriorityToNodes(priorityNodes, farmNodes []uint32) []uint32 {
+	updatedFarmNodes := make([]uint32, len(farmNodes))
+	copy(updatedFarmNodes, farmNodes)
+
+	updatedPriorityNodes := make([]uint32, len(priorityNodes))
+	copy(updatedPriorityNodes, priorityNodes)
```
Eslam-Nawara: I think the copy and delete part should be replaced with empty slices and append nodes if exist.

farmerbot/internal/find_node_test.go
```diff
@@ -67,65 +67,71 @@ func TestFindNode(t *testing.T) {
...
-		assert.Contains(t, farmerbot.nodes, nodeID)
 
-		farmerbot.addNode(oldNode1)
-		farmerbot.addNode(oldNode2)
+		_, node, err := farmerbot.getNode(nodeID)
+		assert.NoError(t, err)
+		assert.Contains(t, farmerbot.nodes, node)
+
+		assert.NoError(t, farmerbot.updateNode(oldNode1))
+		assert.NoError(t, farmerbot.updateNode(oldNode2))
```
Eslam-Nawara: Why are we updating the `oldNodes`

rawdaGastan: To reset changes made by `findNode` function

Eslam-Nawara (approved):

farmerbot/internal/farmerbot_test.go
```diff
@@ -32,6 +32,7 @@ func TestFarmerbot(t *testing.T) {
 	inputs := Config{
 		FarmID:        1,
 		IncludedNodes: []uint32{1, 2},
+		PriorityNodes: []uint32{2},
```
xmonader, line 35: I think we need multiple validations for priority nodes field e.g duplicated record, invalid node id, missing node id, node id that doesn't belong to a farm, .. etc

rawdaGastan: To be added: 
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/1017

xmonader (approved):

Issue #1010, opened 2024-05-13
support filtering with node status list


Omarabdul3ziz:
### Is your feature request related to a problem? Please describe

support filtering with list when possible. it is now available for `nodes?farm_ids` and `contracts?state=`

### Describe the solution you'd like

other filters can take a list too like `nodes?status=` and others

Eslam-Nawara:
Verified: 
I queried multiple cases

- status list [up, down] https://gridproxy.dev.grid.tf/nodes?size=99999&status=up,down

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/67752395/7476f3cf-6419-4d15-9c90-7d01228c225e)
the returned nodes had no standby nodes
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/67752395/b0624fd2-aa98-4200-aa6b-b99ae4e2951b)

- status list [standby, down] https://gridproxy.dev.grid.tf/nodes?size=99999&status=up,down

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/67752395/29abd257-8028-4168-a2ae-091d031da95a)

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/67752395/583f4860-80b9-42d3-8696-79e0ee67e18c)

PR #1021, opened 2024-05-20
allow filtering nodes by list of statuses

Omarabdul3ziz:
### Description
allow filtering nodes by list of statuses

### Changes
- change the status field type in node/farm/stats filters to be slice
- update the node status decision for sql condition
- update tests to cover the new changes
- update node status filtering at the other packages

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/1010

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation


ashraffouda (approved):

grid-proxy/pkg/client/grid_client_test.go
```diff
@@ -344,12 +343,11 @@ func testSuccess(t *testing.T, f ProxyFunc) {
 }
 
 func TestPrepareURL(t *testing.T) {
-	status := "st"
 	freeMRU := uint64(10)
 	farmIDs := []uint64{1, 2, 3}
 	dedicated := true
 	filter := types.NodeFilter{
-		Status:    &status,
+		Status:    []string{"st"},
```
xmonader, line 350: Can you please explain this test? I don't get what it does

grid-proxy/tests/queries/utils.go
```diff
@@ -59,3 +59,12 @@ func SerializeFilter[F Filter](f F) string {
 
 	return res
 }
+
+func getRandomSlice[T any](slice []T) []T {
```
xmonader: Where is that used?

xmonader: I also would pass the number of elements that I need in the slice as a parameter with a guard clause. And maybe rename it to `RandomSliceFrom`

grid-proxy/tests/queries/utils.go
```diff
@@ -59,3 +59,12 @@ func SerializeFilter[F Filter](f F) string {
...
+	randomLen := rand.Intn(len(slice))
```
xmonader: I don't see a rand.Seed call?

grid-proxy/pkg/client/grid_client_test.go
Omarabdul3ziz: it validates the output of the `prepareUrl` function. by making sure the `got` url path/query - returned from the `prepareUrl` function - equals to the `want` url path/query - created from a literal string -

so having a filter
```go
types.NodeFilter{
    Status:    []string{"st"}
}
```
should output `?status=st` and so on

Omarabdul3ziz: it validates the output of the `prepareUrl` function. by making sure the `got` url path/query - returned from the `prepareUrl` function - equals to the `want` url path/query - created from a literal string -

so having a filter
```go
types.NodeFilter{
    Status:    []string{"st"}
}
```
should output `?status=st` and so on

grid-proxy/tests/queries/utils.go
Omarabdul3ziz: found it here. it is based on a flag
https://github.com/threefoldtech/tfgrid-sdk-go/blob/3358c61e65e67e157b5ce497ceb9799b86c101da/grid-proxy/tests/queries/main_test.go#L60

Issue #1039, opened 2024-05-27
🐞 [Bug]: integration tests fail


rawdaGastan:
### What happened?

Integration tests fail after RMB panics

https://github.com/threefoldtech/tfgrid-sdk-go/actions/runs/9244156325

### which network/s did you face the problem on?

Dev, QA, Test, Main

rawdaGastan:
Mainnet has a problem of `notEnoughPublicIPs` error (not solved yet)
memory issue (like ssds and hhds)

rawdaGastan:
Working: https://github.com/threefoldtech/tfgrid-sdk-go/actions/runs/9279688159

Issue #1038, opened 2024-05-27
🐞 [Bug]: tfrobot workflow failed due to rmb panic


Eslam-Nawara:
### What happened?

tfrobot nightly workflow failed due to rmb panic
https://github.com/threefoldtech/tfgrid-sdk-go/actions/runs/9249134007/job/25440491988

### which network/s did you face the problem on?

Main

### Version

latest

### Relevant log output

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/67752395/fbe56a36-7afd-4e44-8cc5-8f11531fd75c)

rawdaGastan:
It is working now: 
https://github.com/threefoldtech/tfgrid-sdk-go/actions/runs/9280182204/job/25534022738

PR #1042, opened 2024-05-27
check if channel is closed before writing

rawdaGastan:
### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/1039
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/1038

### DIFF
--- a/rmb-sdk-go/peer/connection.go
+++ b/rmb-sdk-go/peer/connection.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"sync/atomic"
 	"time"
 
 	"github.com/gorilla/websocket"
@@ -31,9 +30,23 @@ type InnerConnection struct {
 }
 
 type send struct {
-	data   []byte
-	err    chan error
-	closed *atomic.Bool
+	data []byte
+	err  chan error
+}
+
+func (s *send) reply(ctx context.Context, err error) error {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Debug().Msgf("recovered from panic: %v", r)
+		}
+	}()
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case s.err <- err:
+		return err
+	}
 }
 
 // Reader is a channel that receives incoming messages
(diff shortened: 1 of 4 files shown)


rmb-sdk-go/peer/connection.go
```diff
@@ -133,10 +140,12 @@ func (c *InnerConnection) loop(ctx context.Context, con *websocket.Conn, output
...
-			select {
-			case sent.err <- err:
-			case <-ctx.Done():
-				return ctx.Err()
+			if !sent.closed.Load() {
+				select {
+				case sent.err <- err:
+				case <-ctx.Done():
+					return ctx.Err()
+				}
```
muhamadazmy: Still potential race condition here. Even if u use atomic bool, what prevents us from seeing that closed is false, then later the send is closed while blocked on the write inside the select statement

Proposed solution is the following:
- Implement a `reply(ctx Context, err error)` function on the send object
- replace this block with a call `send.reply(ctx, err)`
- in the reply implementation you still do 
```
select {
  case <-ctx.Done():
     return ctx.Err()
  case send.err <- err:
```
But you also add a defer recover section.

Any check you gonna make will have a potential race condition anyhow

muhamadazmy (changes requested):

rmb-sdk-go/peer/connection.go
muhamadazmy: Can be better if we include the `send` context from the send itself. Since the send context can be different from the context passed to the loop. You will have to try this out 

so basically  in send
```
send := send {
   ctx: ctx
   err: (channel)
   data: ...
}
```

 in the reply(err error)
```
select {
  // <- self.ctx 
  // <- write error
```

rmb-sdk-go/peer/connection.go
```diff
@@ -34,6 +34,21 @@ type send struct {
 	err  chan error
 }
 
+func (s *send) reply(ctx context.Context, err error) error {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Debug().Err(r.(error)).Send()
```
muhamadazmy: this will panic again. there is absolutely no grantee that the returned "r" is of type error because i can do `panic("hello world")` which will return a string in recover() in this case

muhamadazmy: just format the object as a string instead

muhamadazmy (approved):

Issue #423, opened 2023-10-18
Supporting Filtering Objects by ID List in gridProxy


Mahmoud-Emad:
**Description:**

### Problem Statement:
Currently, gridProxy lacks the ability to filter objects based on a list of specific IDs. This functionality is crucial for efficient data retrieval and manipulation, especially when dealing with large datasets. The absence of such a feature limits the flexibility and usefulness of the proxy.

### Proposed Solution:
Implement the ability to filter objects in gridProxy by a list of data or specific IDs. This will enhance the proxy's functionality and provide users with a powerful tool for querying and managing their data.

### Details:
- **Endpoint:** `/nodes`
- **Filtering Parameter:** `node_id__in=[1,2,3]` (Example)
- **Expected Behavior:** When the filtering parameter `node_id__in` is provided in the URL, the gridProxy should filter the nodes based on the provided list of IDs (in this example, IDs 1, 2, and 3).

The same goes for all fields e.g. if I want to filter with multiple statues it'll be like => `status__in=['up', 'down']`.

xmonader:
While I understand the need, this is already implemented in the graphql. Which story is this needed for? can it be in 3.13?

Mahmoud-Emad:
@xmonader

The usage of it can be in several cases; here is an example.

Imagine if I have deployments on nodes 1, 2, and 3, and I want to check if these nodes are offline or online. Why not get the details of them all at once? So when filtering by their IDs, it'll be better than sending a request to load all offline nodes and then looping over them.

What do you think?

xmonader:
I understand, but still relevant to which story in tfgrid-sdk-ts?

Mahmoud-Emad:
This issue, https://github.com/threefoldtech/tfgrid-sdk-ts/issues/1220, instead of requesting to load all offline nodes and then looping over them to check if the node ID matches the deployment node ID, it'll be better to get only the nodes that I really need.

Omarabdul3ziz:
we already provided filtering by list for some filters like
- `/nodes?farm_ids=1,2`
- `/nodes?status=up,standby`
- `/contracts?state=Created,GracePeriod`

will work on adding it for `/nodes?node_id`

Omarabdul3ziz:
verified the new filter `node_ids` https://gridproxy.dev.grid.tf/nodes?node_ids=11,14,1,9999,16
returns nodes 11,14,16 
and the old filter `node_id` is still that same

PR #1062, opened 2024-06-04
allow filtering by list of node ids

Omarabdul3ziz:
### Description
support filtering nodes by a list of node ids

### Related Issues
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/423

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/grid-proxy/pkg/types/nodes.go
+++ b/grid-proxy/pkg/types/nodes.go
@@ -140,7 +140,8 @@ type NodeFilter struct {
 	Rented            *bool    `schema:"rented,omitempty"`
 	RentedBy          *uint64  `schema:"rented_by,omitempty"`
 	AvailableFor      *uint64  `schema:"available_for,omitempty"`
-	NodeID            []uint64 `schema:"node_id,omitempty"`
+	NodeID            *uint64  `schema:"node_id,omitempty"`
+	NodeIDs           []uint64 `schema:"node_ids,omitempty"`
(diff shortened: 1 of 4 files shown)


grid-proxy/pkg/types/nodes.go
```diff
@@ -140,7 +140,7 @@ type NodeFilter struct {
 	Rented            *bool    `schema:"rented,omitempty"`
 	RentedBy          *uint64  `schema:"rented_by,omitempty"`
 	AvailableFor      *uint64  `schema:"available_for,omitempty"`
-	NodeID            *uint64  `schema:"node_id,omitempty"`
+	NodeID            []uint64 `schema:"node_id,omitempty"`
```
ashraffouda: this should  be `NodeIDs instead to reflect it is a list 
 also consider changing it in `schema:"node_ids,omitempty"``
 and this for all relevant cases in other files
 also I prefer we have another endpoint which accepts `node_ids` in order not to break the api with clients or we handle all clients in this release if we will change the old endpoint

ashraffouda (changes requested):

grid-proxy/pkg/types/nodes.go
Omarabdul3ziz: added a new filter `node_ids` and kept the old one `node_id` as is, to not break any handled part from a client.

ashraffouda (approved):

Issue #1034, opened 2024-05-24
tfrobot workflow been failing while marked as succeeded because of the return code


xmonader:
![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/64129/00c6e03f-d46b-4d4e-9e94-0be7009a9618)

![image](https://github.com/threefoldtech/tfgrid-sdk-go/assets/64129/53659a99-4be5-47d5-9955-73d8005536a4)

PR #1068, opened 2024-06-06
exit with error if any of the deployments failed

Eslam-Nawara:
### Description

tfrobot workflows doesn't fail if the deployments failed

### Changes

update tfrobot to exit with exit(1) if the deployments failed

### Related Issues

- #1034 

### Checklist

- [ ] Tests included
- [x] Build pass
- [ ] Documentation

### DIFF
--- a/tfrobot/pkg/deployer/deployer.go
+++ b/tfrobot/pkg/deployer/deployer.go
@@ -25,9 +25,10 @@ const (
 	maxGoroutinesCount = 100
 )
 
-func RunDeployer(ctx context.Context, cfg Config, tfPluginClient deployer.TFPluginClient, output string, debug bool) error {
+func RunDeployer(ctx context.Context, cfg Config, tfPluginClient deployer.TFPluginClient, output string, debug bool) *multierror.Error {
 	passedGroups := map[string][]*workloads.Deployment{}
-	failedGroups := map[string]string{}
+	var failedGroupsErr *multierror.Error
+
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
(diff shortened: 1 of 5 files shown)


tfrobot/pkg/deployer/deployer.go
```diff
@@ -97,15 +98,20 @@ func RunDeployer(ctx context.Context, cfg Config, tfPluginClient deployer.TFPlug
 	endTime := time.Since(deploymentStart)
 
 	// load deployed deployments
-	outputBytes, err := loadAfterDeployment(ctx, tfPluginClient, passedGroups, failedGroups, cfg.MaxRetries, filepath.Ext(output) == ".json")
-	if err != nil {
-		return err
+	outputBytes, errs := loadAfterDeployment(ctx, tfPluginClient, passedGroups, cfg.MaxRetries, filepath.Ext(output) == ".json")
+	if errs != nil {
+		failedGroupsErr = multierror.Append(failedGroupsErr, errs.Errors...)
```
rawdaGastan, line 103: why we don't return here?

tfrobot/pkg/deployer/load.go
```diff
@@ -23,31 +24,37 @@ import (
...
-		failedGroups[group] = err
+	if errs != nil {
+		failedGroupsErr = multierror.Append(failedGroupsErr, errs.Errors...)
 	}
 
-	outputBytes, err := parseDeploymentOutput(passedGroups, failedGroups, asJson)
+	outputBytes, err := parseDeploymentOutput(passedGroups, asJson)
 	if err != nil {
-		return err
+		failedGroupsErr = multierror.Append(failedGroupsErr, err)
```
rawdaGastan, line 42: same here?

tfrobot/pkg/deployer/deployer.go
Eslam-Nawara: The load fails if some of the deployments failed, so maybe some deployments return, and we need to print their output before returning, and same for the load

rawdaGastan (approved):

Issue #1133, opened 2024-07-28
Add wireguard option to tfrobot


AbdelrahmanElawady:
### Is your feature request related to a problem? Please describe

I can't request wireguard access when deploying VMs using tfrobot

### Describe the solution you'd like

Add the option to do that through the Yaml input file

PR #1139, opened 2024-07-30
added wireguard option to tfrobot

amr-yassir:
### Description
Add wireguard option to tfrobot

### Changes
Add wireguard option to tfrobot

### Related Issues
* [CLI: Add wireguard option to tfrobot #1133](https://github.com/threefoldtech/tfgrid-sdk-go/issues/1133)

### DIFF
--- a/tfrobot/example/test.yaml
+++ b/tfrobot/example/test.yaml
@@ -49,5 +49,17 @@ vms:
     entry_point: /sbin/zinit init
     ssh_key: key1
 
+  - name: example_d #test deployment with wireguard access
+    vms_count: 3
+    node_group: group_a
+    cpu: 1
+    mem: 1
+    public_ip4: true
+    mycelium_ip: true
+    flist: https://hub.grid.tf/tf-official-apps/base:latest.flist
+    entry_point: /sbin/zinit init
+    wireguard: true
+    ssh_key: key1
+
 ssh_keys:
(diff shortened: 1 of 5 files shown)


Eslam-Nawara:
please provide a description to the pr, use the template.

also please add another vm group [here](https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/tfrobot/example/test.yaml) to test deploying vms with wg
>> *Note:* tfrobot is tested nightly with this file so it's considered a test for the robot

tfrobot/example/test.yaml
```diff
@@ -49,5 +49,17 @@ vms:
     entry_point: /sbin/zinit init
     ssh_key: key1
 
+  - name: example_d #test deployment with wireguard access
+    vms_count: 3
+    node_group: group_a
+    cpu: 1
+    mem: 1
+    public_ip4: true
```
rawdaGastan: That won’t work because number of public ips won’t match the requested ones

tfrobot/example/test.yaml
```diff
@@ -16,6 +16,11 @@ node_groups:
     free_cpu: 2
     free_mru: 8
 
+  - name: group_d
+    nodes_count: 2
+    free_cpu: 2
+    free_mru: 16
```
rawdaGastan: FYI: you can request less mru because you are using only 2 vms with 1 memroy (cpu too)

rawdaGastan (approved):

Issue #1158, opened 2024-08-20
Use the same master's filst when adding a new worker to kubernetes cluster


AhmedHanafy725:
https://github.com/threefoldtech/tf-images/issues/292

Eslam-Nawara:
you mean in the cli? @AhmedHanafy725

AhmedHanafy725:
> you mean in the cli? @AhmedHanafy725

everywhere we use kubernetes

Omarabdul3ziz:
couldn't add a new worker to a cluster with cli
![image](https://github.com/user-attachments/assets/09cb275f-c4c1-4a00-ac76-9a1d4947aacd)

Eslam-Nawara:
### Work completed
- After updating k8s node to be a normal vm, this change introduced multiple problems to updating the cli.
- needed to add the network name to all workers
- workers name were not updating right
- loading the cluster didn't load the env vars of the cluster

worked on fixing all the issue and the update is working fine now

Omarabdul3ziz:
verified. new worker added successfully, used the master flist and the env vars returned in the response

PR #1162, opened 2024-08-21
update k8s clusters to create all nodes using the same flist

Eslam-Nawara:
### Description

update k8s clusters to create all nodes using the same flist

### Changes

update k8s clusters to create all nodes using the same flist

### Related Issues

- #1158 

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

### Tests:
- `Grid client integration tests`: https://github.com/threefoldtech/tfgrid-sdk-go/actions/runs/10680266593


grid-client/workloads/k8s.go
```diff
@@ -197,6 +202,33 @@ func (k *K8sCluster) ValidateIPranges() error {
 	}
 
 	for _, w := range k.Workers {
+		// // ValidateFlist validate check sums for k8s flist
```
rawdaGastan: shouldn't it be removed?

grid-client/workloads/k8s.go
```diff
@@ -298,3 +305,18 @@ func (k *K8sNode) zosWorkload(cluster *K8sCluster, isWorker bool) (K8sWorkloads
 
 	return K8sWorkloads
 }
+
+func (k *K8sCluster) validateFlist() error {
```
rawdaGastan: I think we validate the flists for master and workers, this function assigns the flist

Eslam-Nawara: it also validate that either the cluster flist or the master flist exists

rawdaGastan: I think we should validate the cluster flist only, others are validated,
Also we should display warning for example if cluster, master and workers flists don't match, what do you think?

rawdaGastan (approved):

Issue #1187, opened 2024-09-05
TFRobot Add Mycelium in all examples


mik-tf:
### Is your feature request related to a problem? Please describe

# Situation

- We want to have Mycelium network in all TF examples
- Now there is Mycelium for yaml but not json:
  - https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/tfrobot/example/conf.yaml
  - https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/tfrobot/example/conf.json 

### Describe the solution you'd like

# Todo

- Add mycelium in all examples (json)

# References

- As asked by @xmonader in info_grid: https://github.com/threefoldtech/info_grid/issues/617

PR #1188, opened 2024-09-05
Update conf.json with mycelium as true

mik-tf:
# Related Issue

- #1187 

# Work Done

- Added mycelium as true for TFRobot json example


rawdaGastan:
Already updated
- https://github.com/threefoldtech/tfgrid-sdk-go/pull/1197

Issue #1209, opened 2024-09-25
Add queries for zos4 on nodes/farms endpoint


AhmedHanafy725:
### Is your feature request related to a problem? Please describe

some networks are not supported on zos4 so we need queries to filter them.

### Describe the solution you'd like

add queries for the node if it supports:
- `wireguard`
- `yggdrasil`
- `public ips`

Also, please add them as fields on the node object

same queries should be added too to the farms' queries

ashraffouda:
@Eslam-Nawara we have features now which provide these info, right?
plz add docs on how to use it and link it in the issue

Omarabdul3ziz:
after the discussion, i think having a filter `features` and `node_features` which takes a list of features and filter the nodes that support it will be more suitable for the usecase. 

- `/nodes?features=network,zmachine,zdb`
- `/nodes?features=zmachine-light,network-light`

and in the node response will add the full list of the supported features

no need for wg/ygg/pubIp filters in that case, knowing that network-light nodes doesn't support them. 

is that okay?

rawdaGastan:
### Verifications

- Tried https://gridproxy.dev.grid.tf/nodes?features=zmachine-light,network-light
- Got nodes for light deployments


- Tried https://gridproxy.dev.grid.tf/nodes?features=network,zmachine,zdb
- Got nodes for v3 deployments

- Tried https://gridproxy.dev.grid.tf/farms?node_features=zmachine-light,network-light
- Got farms supports v4 nodes

### Notes

- I found a node with `ipv4` in ight filter
- Not all nodes supports mycelium in light filter too

Omarabdul3ziz:
i suspect it is because some nodes missed the features set update. 


checking https://gridproxy.dev.grid.tf/nodes?features=zmachine-light&status=up only the up nodes (2 now) they both have mycelium and not ipv4 as expected

the other down nodes are
- 1 node in freefarm which last update was 14 days which is before the update
- 3 nodes are a dev nodes. mostly run from source code which hadn't have the update

PR #1214, opened 2024-09-29
init the features indexer

Omarabdul3ziz:
- modify the node response to have the supported features
- add the features indexer to the list of indexers health
- support filtering nodes based on its features
- support filtering farms based on its nodes' features
- add the indexer work that calls the node and decide its have a light network version or the normal one
- decide a full day as the interval for the features indexer
- update swagger docs with the new filters
- add random value generators for the new filters
- add the test conditions for the new filter tests
- add the new table to schema and add a dump generator and loader
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/1209

### Checklist

- [x] Tests included
- [x] Build pass
- [x] Documentation


ashraffouda (approved):

xmonader:
image

Can we discuss? Seems that we only care here about wg, ygg, public ip only? afaik there're changes for other workloads

I was expecting that we will have the version and the node features (all of them) indexed somehow

xmonader (approved):

Issue #1213, opened 2024-09-29
Support zos4 deployments


rawdaGastan:
### Describe the solution you'd like

zos4 now have different workload typs than zos3 
it has networklight instead of network amd zmachine_light instead of zmachine
  - [network light](https://github.com/threefoldtech/zos/blob/zos-light/pkg/gridtypes/zos/network_light.go)
  - [zmachine light](https://github.com/threefoldtech/zos4/blob/main/pkg/gridtypes/zos/network_light.go)

also zos4 nodes doesn't support wireguard, ygg and public config stuff

rawdaGastan:
- You can try to deploy using this script: 
- https://github.com/threefoldtech/tfgrid-sdk-go/blob/c59cf061a57a7b793c8e2d9034a30b5c8401fcb5/grid-client/scripts/deploy_single_vm_with_mycelium_light/main.go
- still under testing

Nabil-Salah:
# Verfication steps:
I tried to run this script: https://github.com/threefoldtech/tfgrid-sdk-go/blob/c59cf061a57a7b793c8e2d9034a30b5c8401fcb5/grid-client/scripts/deploy_single_vm_with_mycelium_light/main.go 
this script use [network light](https://github.com/threefoldtech/zos/blob/zos-light/pkg/gridtypes/zos/network_light.go) and [zmachine light](https://github.com/threefoldtech/zos4/blob/main/pkg/gridtypes/zos/network_light.go)

# It worked fine with zos4 machine with id: 264, 259

Listed on dashboard:

![image](https://github.com/user-attachments/assets/4e2c5e37-b3d9-43f2-b0ba-f8a5c26f97a2)

With Output:

![image](https://github.com/user-attachments/assets/01a3e77b-a4cd-4053-80f1-d42c0eed5246)

![image](https://github.com/user-attachments/assets/9f5d1de0-fed3-4cc4-ae92-e76039548ae1)

# I tried to ping6 mycelium IP and was fine

![image](https://github.com/user-attachments/assets/b6a549c2-697d-4866-851d-4a0b7981c2f4)

it fully addresses the issue linked
there are tests for VM light and it passes
the new code is documented well

PR #1224, opened 2024-10-15
support cli deployment light

rawdaGastan:
### Description

- If no `ipv4`, `ipv6` nor `ygg`flags are given then we will deploy a light network and vm

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/1213

### example

```console
$ tfcmd deploy vm --name examplevm -ssh ~/.ssh/id_rsa.pub --cpu 2 --memory 4 --disk 10
12:06PM INF starting peer session=tf-1508255 twin=192
12:06PM INF deploying network
12:06PM INF deploying vm
12:07PM INF vm mycelium ip: 530:d5c2:2f7a:cae4:ff0f:a4c2:822a:ac24
```

image


grid-cli/cmd/deploy_vm.go
```diff
@@ -152,40 +128,38 @@ var deployVMCmd = &cobra.Command{
...
-				cmd.Context(),
-				t,
-				filter,
-				ssd,
-				nil,
-				rootfss,
-			)
-			if err != nil {
-				log.Fatal().Err(err).Send()
+		// if no public ips or yggdrasil then we should go for the light deployment
```
xmonader, line 132: We should go for the light one first, but if there're no enough nodes found to deploy, we should deploy the normal Vm

grid-cli/cmd/deploy_vm.go
```diff
@@ -152,40 +130,48 @@ var deployVMCmd = &cobra.Command{
...
+				Flist:          flist,
+				Entrypoint:     entrypoint,
+				MyceliumIPSeed: seed,
+			}
+			err = executeVMLight(cmd.Context(), t, vm, node, farm, disk, volume)
+			if err == nil {
+				return nil
+			}
+
+			if err != nil && !errors.Is(err, errNoNodesFound) {
```
Eslam-Nawara: the error is guaranteed to not be nil here so you need just to only check if `!errors.Is(err, errNoNodesFound)`

grid-cli/cmd/deploy_vm.go
```diff
@@ -223,7 +209,116 @@ func init() {
...
+		nodes, err := deployer.FilterNodes(
+			ctx,
+			t,
+			filter,
+			ssd,
+			nil,
+			rootfss,
+		)
+		if err != nil {
+			return errors.Wrap(errNoNodesFound, err.Error())
```
Eslam-Nawara: deployer.FilterNodes already returns `ErrNoNodesMatchesResources` if no nodes found to match the required resources, I think you should use it instead.
check https://github.com/threefoldtech/tfgrid-sdk-go/blob/a5f890d80a127d88c4c2863bb5e91d6d9cedbd5d/grid-client/deployer/node_filter.go#L24

Eslam-Nawara (approved):

PR #1229, opened 2024-10-20
stick with stdlib errors instead of custom ones

Omarabdul3ziz:
- stick with stdlib errors instead of custom ones.
- add timeout error check for the returned errors from http.Client while shuffling stack endpoints

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation


grid-proxy/pkg/client/grid_client.go
```diff
@@ -387,6 +387,13 @@ func (g *Clientimpl) PublicIps(ctx context.Context, filter types.PublicIpFilter,
 	return ips, totalCount, nil
 }
 
+func isTimeoutError(err error) bool {
+	if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
```
xmonader, line 391: shouldn't we avoid that and use errors.As instead?

```
func isTimeoutError(err error) bool {
	var netErr net.Error

	if errors.As(err, &netErr) && netErr.Timeout() {
		return true
	}
	return errors.Is(err, context.DeadlineExceeded) || 
	       strings.Contains(err.Error(), "timeout awaiting response headers")
}
```

Also we can check against this one first https://pkg.go.dev/net/url#Error

xmonader (approved):

grid-proxy/pkg/client/grid_client.go
Omarabdul3ziz: actually, we will need to type assert the error again to access `.Timeout()`
I reverted to using type assertion and added a check for `url.Error`

xmonader (approved):

PR #1306, opened 2025-01-12
create zos4 registerer server

Eslam-Nawara:
### Description

create zos4 registerer server

### Changes

### Related Issues

- threefoldtech/tfgrid4-sdk-go#20 

### Checklist

- [ ] Tests included
- [x] Build pass
- [x] Documentation


ashraffouda (approved):

node-registrar/cmds/main.go
```diff
@@ -0,0 +1,74 @@
...
+	flag.StringVar(&f.SSLMode, "ssl-mode", "disable", "postgres ssl mode[disable, require, verify-ca, verify-full]")
+	flag.IntVar(&sqlLogLevel, "sql-log-level", 2, "sql logger level")
+
+	flag.BoolVar(&f.version, "v", false, "shows the package version")
+	flag.BoolVar(&f.debug, "debug", false, "allow debug logs")
+	flag.IntVar(&f.port, "server-port", 443, "server port")
+	flag.StringVar(&f.domain, "domain", "", "domain on which the server will be served")
+
+	flag.Parse()
+	f.SqlLogLevel = logger.LogLevel(sqlLogLevel)
```
xmonader: Do bound check on the loglevel before setting it https://github.com/go-gorm/gorm/blob/master/logger/logger.go#L39

node-registrar/cmds/main.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		log.Fatal().Msg("failed to open database with the specified configurations")
+	}
+
+	s, err := server.NewServer(db)
+	if err != nil {
+		log.Fatal().Msg("failed to start gin server")
+	}
+
+	err = s.Router.Run(fmt.Sprintf("%s:%d", f.domain, f.port))
```
xmonader: Shouldn't be ble to access the Router, replace with

if err := s.Run(.....); err!=nil {
   log fatal here

}

node-registrar/cmds/main.go
```diff
@@ -0,0 +1,74 @@
...
+	flag.StringVar(&f.domain, "domain", "", "domain on which the server will be served")
+
+	flag.Parse()
+	f.SqlLogLevel = logger.LogLevel(sqlLogLevel)
+
+	if f.version {
+		log.Info().Str("version", version).Str("commit", commit).Send()
+		return
+	}
+	if f.domain == "" {
```
xmonader: No validation needed on the value passed here?

xmonader: I noticed rest of the fields aren't validated, e.g ports

node-registrar/cmds/main.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	zerolog.SetGlobalLevel(zerolog.InfoLevel)
+	if f.debug {
+		zerolog.SetGlobalLevel(zerolog.DebugLevel)
+	}
+
+	db, err := db.NewDB(f.Config)
+	if err != nil {
+		log.Fatal().Msg("failed to open database with the specified configurations")
```
xmonader: utilize the error.

node-registrar/cmds/main.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	s, err := server.NewServer(db)
+	if err != nil {
+		log.Fatal().Msg("failed to start gin server")
```
xmonader: utilize the err

node-registrar/cmds/main.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	err = s.Router.Run(fmt.Sprintf("%s:%d", f.domain, f.port))
+	if err != nil {
+		log.Fatal().Msg("failed to run gin server")
```
xmonader: utilize the err

node-registrar/main.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+import "github.com/threefoldtech/tfgrid-sdk-go/node-registrar/cmds"
+
+func main() {
```
xmonader: move this to cmds/registrar.go

node-registrar/cmds/main.go
```diff
@@ -0,0 +1,74 @@
+package cmds
+
```
xmonader: this file should be registar.go and should have the main func

func main() {

    ...
    if err:=Run() {
        log.Fatal(....)
    }
}
and change the Run function to return an err

node-registrar/pkg/db/db.go
```diff
@@ -0,0 +1,59 @@
...
+
+type Config struct {
+	Host        string
+	Port        int
+	DBName      string
+	User        string
+	Password    string
+	SSLMode     string
+	SqlLogLevel logger.LogLevel
+	MaxConns    int
```
xmonader: use a proper type e.g uint and make sure that's within the bounds

node-registrar/pkg/db/db.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/pkg/errors"
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
+)
+
+type Config struct {
+	Host        string
+	Port        int
```
xmonader: use the appropriate type e.g uint

node-registrar/pkg/db/db.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return db, errors.Wrapf(err, "Failed to connect to the database: %v", err)
+	}
+
+	sql, err := gormDB.DB()
+	if err != nil {
+		return db, errors.Wrap(err, "failed to configure DB connection")
+	}
+
+	sql.SetMaxIdleConns(3)
```
xmonader: idle connections 3 -> use that as a const

node-registrar/pkg/db/db.go
```diff
@@ -0,0 +1,59 @@
...
+	sql.SetMaxOpenConns(c.MaxConns)
```
xmonader: change the name to MaxOpenConnections

node-registrar/pkg/db/db.go
```diff
@@ -0,0 +1,59 @@
...
+
+// PostgresDatabase postgres db client
+type DataBase struct {
+	gormDB     *gorm.DB
+	connString string
+}
+
+var ErrRecordNotFound = errors.New("could not find any records")
+
+func NewDB(c Config) (db DataBase, err error) {
```
xmonader: I would try to split the concerns of the function, and ensure I do a ping on the DB when done

node-registrar/pkg/db/db.go
```diff
@@ -0,0 +1,59 @@
...
+	DBName      string
+	User        string
+	Password    string
+	SSLMode     string
+	SqlLogLevel logger.LogLevel
+	MaxConns    int
+}
+
+// PostgresDatabase postgres db client
+type DataBase struct {
```
xmonader: Name should Be `Database` instead of `DataBase`

node-registrar/cmds/main.go
sameh-farouk: The db resources are not handled properly. Consider Implementing a graceful shutdown and make sure to close database connections and finish ongoing requests properly

node-registrar/cmds/main.go
```diff
@@ -0,0 +1,74 @@
...
+	flag.IntVar(&f.Config.Port, "postgres-port", 5432, "postgres port")
+	flag.StringVar(&f.DBName, "postgres-db", "", "postgres database")
+	flag.StringVar(&f.User, "postgres-user", "", "postgres username")
+	flag.StringVar(&f.Password, "postgres-password", "", "postgres password")
+	flag.StringVar(&f.SSLMode, "ssl-mode", "disable", "postgres ssl mode[disable, require, verify-ca, verify-full]")
+	flag.IntVar(&sqlLogLevel, "sql-log-level", 2, "sql logger level")
+
+	flag.BoolVar(&f.version, "v", false, "shows the package version")
+	flag.BoolVar(&f.debug, "debug", false, "allow debug logs")
+	flag.IntVar(&f.port, "server-port", 443, "server port")
```
sameh-farouk: The default port is set to 443 but this is not a conventional choice since the codebase uses a plain HTTP
I would recommend changing the default port to 8080 so it avoids conflicts with other web services and avoids the need for elevated privileges (for Ports below 1024)

node-registrar/pkg/db/db.go
sameh-farouk: Consider making this configurable via the Config struct, similar to MaxConns?

node-registrar/pkg/db/models.go
```diff
@@ -0,0 +1,59 @@
+package db
+
+type Farm struct {
+	FarmID      uint64 `gorm:"primaryKey;autoIncrement" json:"farm_id"` // Primary key
+	FarmName    string `gorm:"size:100;not null;unique" json:"farm_name"`
+	TwinID      uint64 `json:"twin_id" gorm:"not null"`
+	Dedicated   bool   `json:"dedicated"`
+	FarmFreeIps uint64 `json:"farm_free_ips"`
```
sameh-farouk, line 17: We are only storing the number of IP addresses. Where are the actual IP addresses stored or managed?

node-registrar/pkg/db/models.go
```diff
@@ -0,0 +1,59 @@
...
+
+	Nodes []Node `gorm:"foreignKey:farm_id;constraint:OnDelete:CASCADE" json:"nodes"`
+}
+
+type Node struct {
+	NodeID      uint64      `gorm:"primaryKey;autoIncrement" json:"node_id"`
+	FarmID      uint64      `gorm:"not null" json:"farm_id"`
+	TwinID      uint64      `json:"twin_id" gorm:"not null"`
```
sameh-farouk: Where will the twin information be stored? The registrar should also manage creating and updating twins, correct?

node-registrar/pkg/server/routes.go
```diff
@@ -0,0 +1,25 @@
...
+	_ "github.com/threefoldtech/tfgrid-sdk-go/node-registrar/docs"
+
+	"github.com/swaggo/files"
+	"github.com/swaggo/gin-swagger"
+)
+
+func (s *Server) SetupRoutes() {
+	s.Router.GET("/swagger/*any", ginSwagger.WrapHandler(swaggerFiles.Handler))
+
+	farmRoutes := s.Router.Group("farms")
```
sameh-farouk: Consider adding a version prefix to the route groups `/v1`.

node-registrar/pkg/server/routes.go
```diff
@@ -0,0 +1,25 @@
...
+	farmRoutes.GET("/", s.listFarmsHandler)
+	farmRoutes.GET("/:farm_id", s.getFarmHandler)
+	farmRoutes.POST("/", s.createFarmHandler)
+	farmRoutes.PATCH("/", s.updateFarmsHandler)
```
sameh-farouk: Reviewing the implementation of the update farm handler and DB, I see it replaces the existing resource with what we send or creates a new resource if one doesn't already exist. To me, this behavior resembles a `PUT` request (full update) rather than a `PATCH` request (partial update).

If you intend to perform a partial update (meaning that specific fields are updated while the rest remain unchanged) you should modify the implementation accordingly. Also, `Patch` shouldn't create resources. Alternatively, if a full update is indeed your goal, then please consider using the `PUT` method here to reflect that intention.

sameh-farouk: Consider using the path `/farm/:farm_id` for updates as it better aligns with RESTful conventions.

node-registrar/pkg/server/routes.go
```diff
@@ -0,0 +1,25 @@
...
+
+	nodeRoutes := s.Router.Group("nodes")
+	nodeRoutes.GET("/", s.listNodesHandler)
+	nodeRoutes.GET("/:node_id", s.getNodeHandler)
+	nodeRoutes.POST("/", s.registerNodeHandler)
```
sameh-farouk, line 27: Is it intentional that updating nodes is not supported?

node-registrar/pkg/db/models.go
```diff
@@ -0,0 +1,59 @@
...
+	Features    []string    `gorm:"type:jsonb;serializer:json"`
+	PriceUsd    float64     `json:"price_usd"`
+	ExtraFee    uint64      `json:"extra_fee"`
+	Status      string      `gorm:"size:50" json:"status"`
+	Healthy     bool        `json:"healthy"`
+	Dedicated   bool        `json:"dedicated"`
+	Rented      bool        `json:"rented"`
+	Rentable    bool        `json:"rentable"`
+	Uptime      Uptime      `json:"uptime"`
+	Consumption Consumption `json:"consumption"`
```
sameh-farouk: I see that Consumption is a type of string. What does it store? Shouldn't it represent bandwidth consumption?

node-registrar/pkg/db/models.go
```diff
@@ -0,0 +1,59 @@
...
+
+type Node struct {
+	NodeID      uint64      `gorm:"primaryKey;autoIncrement" json:"node_id"`
+	FarmID      uint64      `gorm:"not null" json:"farm_id"`
+	TwinID      uint64      `json:"twin_id" gorm:"not null"`
+	Features    []string    `gorm:"type:jsonb;serializer:json"`
+	PriceUsd    float64     `json:"price_usd"`
+	ExtraFee    uint64      `json:"extra_fee"`
+	Status      string      `gorm:"size:50" json:"status"`
+	Healthy     bool        `json:"healthy"`
```
sameh-farouk: I see that the node model lacks essential information, such as resources. Is this due to new design or requirements? also, node location, certification (unless no longer used in ZOS4)...
Have a look at the current node representation on chain to have a better idea of what would be missing
https://github.com/threefoldtech/tfchain/blob/77e684bcb1b2107087810d9ee4c93b87163cc02a/substrate-node/pallets/pallet-tfgrid/src/migrations/types.rs#L28-L46

node-registrar/pkg/db/models.go
```diff
@@ -0,0 +1,59 @@
...
+	Dedicated   bool        `json:"dedicated"`
+	Rented      bool        `json:"rented"`
+	Rentable    bool        `json:"rentable"`
```
sameh-farouk: We might be trying to track too many aspects here and should instead consider separating data between what represents the node (rarely changed) and the often-volatile state parameters such as healthy, rented, and rentable. This way continuous changes to the state don’t require altering the node table.

node-registrar/pkg/db/models.go
```diff
@@ -0,0 +1,59 @@
...
+	Uptime      Uptime      `json:"uptime"`
```
sameh-farouk, line 38: You are currently storing only the latest uptime report from nodes. How this would be useable? Shouldn't instead support storing multiple uptime records for the same node over time. In this case, it would be more appropriate to use a separate table specifically for uptime reports.

node-registrar/pkg/server/handlers.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{
+		"message": "Farm created successfully",
+		"farm":    farm,
+	})
+}
```
sameh-farouk, line 117: The primary issues with these handlers are as follows:

- Missing API Authorization and Authentication: There is no mechanism to restrict or control access, allowing anyone to perform actions on any resource. For example, one user can update another user's farm without restrictions.
- Lack of Validation: The handlers do not validate input or handle missing fields, which increases the likelihood of invalid or incomplete data being stored in the database.

node-registrar/pkg/server/handlers.go
```diff
@@ -0,0 +1,336 @@
...
+		return
+	}
+
+	err := s.DB.CreateFarm(farm)
+	if err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{
```
sameh-farouk: Use proper HTTP status codes 201 for creation

node-registrar/pkg/server/handlers.go
```diff
@@ -0,0 +1,336 @@
...
+		return
+	}
+
+	err := s.DB.RegisterNode(node)
+	if err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		return
+	}
+
+	c.JSON(http.StatusOK, gin.H{
```
sameh-farouk: Use proper HTTP status codes 201 for creation

node-registrar/pkg/db/models.go
sameh-farouk: As with Uptime, since storing the last report only would be almost useless since the report tracks the usages per specific time window, it would be more appropriate to use a separate table specifically for consumption reports to track multiple consumption reports for the same node over time.

node-registrar/pkg/db/models.go
```diff
@@ -0,0 +1,59 @@
...
+
+type Farm struct {
+	FarmID      uint64 `gorm:"primaryKey;autoIncrement" json:"farm_id"` // Primary key
+	FarmName    string `gorm:"size:100;not null;unique" json:"farm_name"`
+	TwinID      uint64 `json:"twin_id" gorm:"not null"`
+	Dedicated   bool   `json:"dedicated"`
+	FarmFreeIps uint64 `json:"farm_free_ips"`
+
+	Nodes []Node `gorm:"foreignKey:farm_id;constraint:OnDelete:CASCADE" json:"nodes"`
+}
```
sameh-farouk, line 23: Include `CreatedAt` and `UpdatedAt` in your models

sameh-farouk:
Thank you, Eslam. Overall, this is a good beginning. I understand that the requirements are not very clear at the moment, but I believe you should take a closer look at how nodes, farms, and twins are validated, stored, and managed on TFChain.
The [pallet-tfgrid](https://github.com/threefoldtech/tfchain/blob/development/substrate-node/pallets/pallet-tfgrid/readme.md) is the runtime module responsible for that.

node-registrar/pkg/db/models.go
Eslam-Nawara: will need to discuss if we should store the ips here in the server or it should be managed by the node

node-registrar/cmds/main.go
Eslam-Nawara: changed the default to port 8080

node-registrar/pkg/db/models.go
Eslam-Nawara: for now the registrar is not handling twin management, it is using the already created twins, we extend the registrar to create/manage its own twins and the specs of the twins

node-registrar/pkg/server/routes.go
Eslam-Nawara: I updated the behavior to allow update for only few fields of the farm and it doesn't create one if not existing anymore

node-registrar/pkg/server/routes.go
Eslam-Nawara: it was not requested, we can add it if needed

node-registrar/pkg/db/models.go
Eslam-Nawara: updated it to be `[]substrate.NruConsumption` we can update that according to the specs, but I see that it's what the node sends on registration

node-registrar/pkg/db/models.go
Eslam-Nawara: I updated the node and added more information, but we can change it even more as needed

node-registrar/pkg/server/handlers.go
Eslam-Nawara: I added auth middleware to check if the twin sending the request is the same as the node owner, but might need to discuss how twins are represented/managed and update the auth accordingly 

need to add more validation too but didn't get the time to do so

Issue #1346, opened 2025-03-27
"americas" region returns no nodes


scottyeager:
![Image](https://github.com/user-attachments/assets/076cdda8-d58e-4aad-b5b8-34dcae1d6b1d)

Omarabdul3ziz:
previously we had inconsistencies between the data sources used to get region/country and coordinates, so we unified them in this change: https://github.com/threefoldtech/tfgrid-sdk-go/commit/0c0246a1b86c1dfc5804383b372d816f219f38bc

for the region, we used the continent as a fallback since the dataset lacks a specific region field. This worked for all continents except America, where nodes were indexed as either North or South America
![Image](https://github.com/user-attachments/assets/d6b44015-566e-43c6-9522-946786fa53f2)
here https://gridproxy.grid.tf/nodes?region=North%20America

The region filter used exact matching, so your request didn’t work. The mentioned PR changes this to allow subtext matching, so filtering by "America" now includes both continents.

scottyeager:
Thanks for the explanation @Omarabdul3ziz.

I should have included the fact that I expected this filter to work based on this line from the tfrobot README:

https://github.com/threefoldtech/tfgrid-sdk-go/blob/aeab47fd2d819de873266bc9e552003d9d9c9c93/tfrobot/README.md?plain=1#L39

Seems the README should also be updated.

Omarabdul3ziz:
updated and added to swagger docs in proxy

PR #1352, opened 2025-04-13
refactor region filtering

Omarabdul3ziz:
- improve region filtering with case-insensitive matching in database queries
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/1346

### DIFF
--- a/grid-proxy/tests/queries/farm_test.go
+++ b/grid-proxy/tests/queries/farm_test.go
@@ -62,37 +62,18 @@ var farmFilterRandomValueGenerator = map[string]func(agg FarmsAggregate) interfa
 		return &country
 	},
 	"Region": func(agg FarmsAggregate) interface{} {
-		c := changeCase(agg.regions[rand.Intn(len(agg.regions))])
-		if len(c) == 0 {
+		value := getRandomItemSubstring(agg.regions)
+		if len(value) == 0 {
 			return nil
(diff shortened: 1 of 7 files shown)


grid-proxy/tests/queries/farm_test.go
```diff
@@ -67,6 +67,16 @@ var farmFilterRandomValueGenerator = map[string]func(agg FarmsAggregate) interfa
 			return nil
 		}
 
+		runesList := []rune(c)
```
xmonader: what does this block of code do?

grid-proxy/tests/queries/node_test.go
```diff
@@ -130,6 +130,16 @@ var nodeFilterRandomValueGenerator = map[string]func(agg NodesAggregate) interfa
 			return nil
 		}
 
+		runesList := []rune(region)
```
xmonader: seems like that's also duplicated from the code above?

Omarabdul3ziz: yes, it is duplicated same as CityContains and CountryContains filters for both nodes/farms.

basically this snippet tries to get a substring of a random city/country/region from the aggregated values from the database, to pass it the filter and make sure it can search with any part of the value

xmonader (approved):

Issue #1355, opened 2025-04-16
allow enable-disable sentry


rawdaGastan:
### Is your feature request related to a problem? Please describe

sentry returns server errors

### Describe the solution you'd like

Allow users to disable sentry

scottyeager:
I think part of the problem here is that we have Sentry's debug mode always enabled:

https://github.com/threefoldtech/tfgrid-sdk-go/blob/3786ea6724fee9c3caca720d650592831fc89394/grid-client/deployer/sentry.go#L19-L22

That seems to be the reason that there's any Sentry related output on the console. To me it would be logical to have Sentry debug mode disabled by default and toggled on by a flag or environment variable.

PR #1356, opened 2025-04-16
allow to diable sentry in client-cli-tfrobot

rawdaGastan:
### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/1355


xmonader:
Can we make the option disabled by default so the user manually enables it?

rawdaGastan:
@xmonader we agreed on leaving it as is (if we wanted to change it we can in other pr)

Omarabdul3ziz (approved):

PR #1362, opened 2025-05-01
ignore go.work and go.work.sum files

Omarabdul3ziz:
go.work.sum keeps changing even when no dependencies are updated. this leads to noisy diffs and unnecessary commits

according to the [official Go proposal](https://go.googlesource.com/proposal/+/master/design/45713-workspace.md#preventing-files-from-being-checked-in-to-repositories) go.work and go.work.sum are meant for local setups only and should not be checked into version control


xmonader:
@Omarabdul3ziz only @rawdaGastan can approve that. It took us a while to have a proper setup for this mono
repo

rawdaGastan:
I don't think we can ignore them. Some directories depend on others

Issue #1284, opened 2024-12-08
return all node internet speeds


Omarabdul3ziz:
### Is your feature request related to a problem? Please describe

currently, we only return the first non-zero speed from the node speeds (tcp/udp - v4/v6). https://github.com/threefoldtech/tfgrid-sdk-go/blob/22c53179873fd1585b5b0996be0082995bf0b3e9/grid-proxy/internal/indexer/speed.go#L101

### Describe the solution you'd like

update the table schema and api type to include the four speeds

PR #1364, opened 2025-05-06
Return all node speeds

Nabil-Salah:
### Description

Describe the changes introduced by this PR and what does it affect

### Changes

Speed type struct `grid-proxy/pkg/types/indexer.go`

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-go/issues/1284

### Checklist

- [x] Tests included
- [x] Build pass
- [ ] Documentation

image


grid-proxy/internal/indexer/speed.go
```diff
@@ -95,14 +178,21 @@ func parseSpeed(res TaskResult, twinId uint32) (types.Speed, error) {
 		return speed, err
 	}
 
-	// TODO: better parsing
-	// we have four speeds tcp/udp for ipv4/ipv6.
-	// now, we just pick the first non-zero
 	for _, report := range iperfResults {
-		if report.DownloadSpeed != 0 {
-			speed.Download = report.DownloadSpeed
+		isIpv4 := isValidIpv4(report.NodeIpv4)
```
Omarabdul3ziz: can you check this https://pkg.go.dev/net#ParseIP https://pkg.go.dev/net#IP.To4 instead of implementing the validation

grid-proxy/internal/indexer/speed.go
```diff
@@ -95,14 +178,21 @@ func parseSpeed(res TaskResult, twinId uint32) (types.Speed, error) {
...
+			speed.Download = report.DownloadSpeed
+		} else if report.TestType == "udp" && isIpv4 {
+			speed.UDPUploadIPv4 = report.UploadSpeed
+			speed.UDPDownloadIPv4 = report.DownloadSpeed
+		} else if report.TestType == "tcp" && isIpv6 {
+			speed.TCPUploadIPv6 = report.UploadSpeed
+			speed.TCPDownloadIPv6 = report.DownloadSpeed
+		} else if report.TestType == "udp" && isIpv6 {
+			speed.UDPUploadIPv6 = report.UploadSpeed
+			speed.UDPDownloadIPv6 = report.DownloadSpeed
```
Omarabdul3ziz: worth noting that usually the node runs the iperf test against multiple public nodes (for examples dev nodes has two iperf servers on 11/14) 

could we got the average for each connection type between nodes?

grid-proxy/internal/explorer/converters.go
```diff
@@ -69,8 +69,14 @@ func nodeFromDBNode(info db.Node) types.Node {
 			Baseboard: info.Baseboard,
 		},
 		Speed: types.Speed{
-			Upload:   info.UploadSpeed,
-			Download: info.DownloadSpeed,
+			Upload:          info.UploadSpeed,
```
Omarabdul3ziz, line 72: make sure swagger docs are updated, `make docs` and commit the files

Omarabdul3ziz (changes requested):
i have small comments

grid-proxy/internal/indexer/speed.go
Omarabdul3ziz: nvm, no need for calc average it is almost the same. but at least could we return once we had all 4 speeds instead of keep overriding.

also consider some server may fails or return 0

Omarabdul3ziz (approved):

Issue #1203, opened 2024-09-17
add rate limiting in proxy


Omarabdul3ziz:

Nabil-Salah:
Work done:

- Implemented rate limiter with variable sliding window technique.
- Used limiter on server endpoint for Action and ProxyAction handle functions.

Nabil-Salah:
Work done:
- Load tested it to make sure it handles requests correctly and at the rate determined

PR #1385, opened 2025-07-08
Rate limiter for the grid-proxy

Nabil-Salah:
### Description

Describe the changes introduced by this PR and what does it affect

### Changes

### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/1203

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

image


sameh-farouk (commented):
Should we first consider using our reverse-proxy rate limiter? Since it’s a simple IP-based limiter that doesn’t require any business logic, configuring it properly could reduce application complexity and is often more reliable than implementing custom rate-limiting code for such a common use case.

That said, I’m also wondering whether this has been tested behind a reverse proxy. The way GetClientIP parses headers may cause all users to be grouped under the same IP address (depending on how the reverse proxy is configured).

Issue #1388, opened 2025-07-15
🐞 [Bug]: RMB - Flaw in Go Peer's Relay Failover and Selection Strategy


sameh-farouk:
### What happened?

 **Description:**

  The Go peer (rmb-sdk-go/peer) has a critical logical flaw in its send function, which is responsible for sending messages to relays. The current implementation fails to provide a robust failover strategy, and its connection prioritization mechanism is non-functional.

  This deviates significantly from the more resilient implementation in the Rust rmb-peer.

**Issues:**

   1. Incorrect Failover Loop: The send function loops N times for N relays, but it calls Choose() in each iteration. Because Choose() is stateless, the loop does not guarantee that each relay is tried exactly once. It can repeatedly try the same failing relay while completely ignoring other healthy ones.

   2. Non-Functional Connection Prioritization: The WeightSlice that drives the Choose() algorithm is initialized once when the peer is created. The send function modifies the Weight of connections on success or failure, but the totals slice used by Choose() is never recalculated. This means the selection probabilities are static and never adapt to real-time network conditions. The weight-adjustment logic is effectively dead code.

   3. Fail-fast: The send function loops N times for N relays, then gives up. Since the connections are self-healing, the send function should instead keep trying to deliver the message till the TTL expires. 

**Impact:**

   * Reduced Reliability: The peer may fail to send a message even if a healthy relay connection is available because it might not be selected by the flawed loop.
   * No Adaptive Routing: The peer does not prioritize healthier connections, leading to potentially higher latency and more transient errors, as it cannot route around flaky relays.

**Comparison to Rust Peer:**

  The Rust rmb-peer handles this correctly:
   * It loops infinitely (until TTL expiry) over all connections using .cycle().
   * It guarantees every connection is tried.
   * It uses a simple and effective prioritization strategy: the last successful connection is moved to the front of the list (swap(index, 0)).
   * It includes a 1-second backoff after a full cycle of failures.

**Suggested Action:**

  The send function and WeightSlice in the Go peer need to be redesigned to implement a resilient failover strategy correctly. This likely involves:
   1. Removing the current WeightSlice logic.
   2. Implementing a loop that iterates through each connection exactly once per cycle.
   3. Adopting a simple and effective prioritization scheme, similar to the Rust peer's "move-to-front" strategy.

**Note regarding tests**

`TestWeightedRandom` is the core test. It does the following:
[...]

PR #1391, opened 2025-07-23
Fix relay failover mechanism 

sameh-farouk:
### Description

The original WeightSlice implementation had a fundamental flaw where it used static weighted-random selection that didn't adapt to relay failures, causing "sticky" relay behavior. Also, it didn't behave as expected. see this [issue](https://github.com/threefoldtech/tfgrid-sdk-go/issues/1388). The cooldown-based relay failover mechanism proposed in this PR is fair, configurable, and thread-safe. It provides improvement over the original weighted random approach and will ensure reliable relay failover in production environments.

- Replaced WeightSlice with CooldownRelaySet that provides fair, adaptive relay ordering 
- Failed relays are penalized with a cooldown period where it would be deprioritized but not excluded 
- All relays get fair chances based on their recent performance
- Fair retry until envelope expiration

Also, some optimization and validation were introduced:
- Bumping the send timeout from 2 to 5 seconds (per relay attempt). The ideal value is always specific to the environment, but 5 seconds IMO is a more conservative and safer starting point than 2 seconds. It strikes a more balanced compromise, favoring stability and resilience in the face of real-world server conditions, while the cost of a 3-second slower reaction to a truly dead connection is acceptable for the vast majority of applications.
- Validate incoming messages' expiration, and prevent further processing if the message has expired already. This saves resources and is considered a good security practice.

Related Issues:
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/1388


xmonader:
I think maybe a more direct behavior could resolve lots of the complexity in the PR

```go

func (p *Peer) send(ctx context.Context, request *types.Envelope) error {

    sendCtx, cancel := context.WithDeadline(ctx,...)
    defer cancel()

    select {
    case con := <-p.healthyConns:
        err := con.send(sendCtx, bytes)
        if err != nil {
            p.penalize(con) 
            return err // Fail fast, now the caller can decide to retry
        }
        p.healthyConns <- con
        return nil
    case <-sendCtx.Done():
        return fmt.Errorf("something something: %w", sendCtx.Err())
    }
}

func (p *Peer) penalize(con *InnerConnection) {
    time.AfterFunc(p.cooldown, func() {
        select {
        case p.healthyConns <- con:
        default: // log that the peer is shutdown
        }
    })
}
```

sameh-farouk:
> I think maybe a more direct behavior could resolve lots of the complexity in the PR
> 

Thanks for the feeedback. I see the logic, but with a few connections available, I’m concerned about scenarios, such as when there is only one available connection, or losing network connectivity temporaliy leading to penalize all connections. In such scenarios, exclusion could lead to unnecessary downtime since there would be no connection left to attempt a send operation until the cooldown expires.

The primary focus of this PR is to improve the reliability of the failover mechanism and IMO, **deprioritizing** failed connections (not excluding) would be more resilient, as it keeps all options available for retries/failover.

I'm not particularly attached to the idea of scoring the connections in this specfic scope (it was already there, so I kept it).
Personally, if simplifications a top priority here, I'd prefer a simple shuffle and loop approach over both original random selection and your suggestion. What really matters to me is ensuring that all connections are retried.

Open to further thoughts!

rmb-sdk-go/peer/relay_retry_set.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	return result
+}
+
+// MarkFailure updates the penalty for the given relay.
+// T must be a pointer type. Pointer value comparison is used.
+func (s *CooldownRelaySet[T]) MarkFailure(relay T, now time.Time) {
+	for i := range s.Relays {
+		if reflect.ValueOf(s.Relays[i].Relay).Pointer() == reflect.ValueOf(relay).Pointer() {
```
xmonader: I'm not a fan, I believe this could be overly simplified but having that in penalized list and the keeping the healthy ones in the healthy list and shuffle over both. trying first from the healthy then from the penalized

rmb-sdk-go/peer/peer.go
```diff
@@ -44,6 +44,15 @@ type peerCfg struct {
 	enableEncryption bool
 	encoder          encoder.Encoder
 	cacheFactory     cacheFactory
+	relayCooldown    time.Duration
+}
+
+// WithRelayCooldown sets the cooldown duration for relay failover and retry logic.
+// If not set, defaults to 10 seconds.
```
rawdaGastan, line 60: I think 10 seconds is much? no?

rmb-sdk-go/peer/peer.go
```diff
@@ -520,26 +543,28 @@ func (d *Peer) send(ctx context.Context, request *types.Envelope) error {
...
+			if err != nil {
+				errs = multierror.Append(errs, err)
+				set.MarkFailure(con, time.Now())
+				continue
+			}
+			set.MarkSuccess(con)
+			return nil
 		}
-		return nil
+		time.Sleep(100 * time.Millisecond)
```
rawdaGastan, line 601: here too, I fear that affects rmb responses

rmb-sdk-go/peer/peer.go
sameh-farouk: No at all. The cool down will affect only the order in which relays would be tried out

rmb-sdk-go/peer/peer.go
sameh-farouk: This is a very brief pause after exhausting all the relays to prevent a busy loop. If removed, and the connections fail immediately (e.g., the machine is not yet connected to the internet), we won’t leave any breathing space for the CPU to perform other tasks.

rmb-sdk-go/peer/relay_retry_set.go
sameh-farouk: I simplified the code by using *InnerConnection directly, and removing reflection-based comparison to improve readability. The code should now be safer and more straightforward to understand, while functionality remains intact.

The current implementation is simpler than maintaining two separate lists. By using atomic operations on the `LastErrorAt` field, we've simplified the code while making it more robust.

Instead of juggling relays between healthy and penalized lists, which would require careful synchronization using a mutex, we now have a single source of truth for each relay's state. The atomic operations ensure (lock-free) thread safety without the overhead of locks during normal operation, and we avoid constantly reallocating slices.

The beauty of this approach is in its simplicity. We can mark a relay as failed or successful with a single atomic store operation, and the `Sorted()` method efficiently organizes the relays based on their current effective state. IMO, this makes the code easier to reason about while being more performant and simpler than the two-list approach.

rawdaGastan (approved):

sameh-farouk:
**Superseded** by: #1437
This PR’s changes have been merged via the unified feature branch in #1437. Merge commit: c753f8dade0eeece636815d081f9806a54900ecc

**Reason**: We consolidated multiple related PRs to resolve conflicts, preserve consistent behavior, and land the full feature set together (tags, TTL safety, cooldown failover, lifecycle/observability, URL validation, tests/docs).

**Action**: Closing this PR as superseded. Thanks for the work and discussion here.

Issue #1389, opened 2025-07-15
🐞 [Bug]: RMB - Relays Failing Initial Connection Are Permanently Excluded


sameh-farouk:
### What happened?

  Description:

  The rmb-sdk-go/peer implementation has a critical bug where relays that fail to establish an initial connection during the peer's startup are permanently excluded from its active connection pool.

   * How it works (incorrectly):
       1. During the Peer's initialization (peer.go:NewPeer), the getRelayConnections function iterates through the list of configured relay URLs.
       2. For each URL, it calls conn.TryConnect().
       3. If conn.TryConnect() fails (meaning the relay is unreachable or unresponsive at that exact moment), that specific InnerConnection object is not added to the successfulConnections
          slice.
       4. Crucially, the InnerConnection.Start() method (which contains the robust, self-healing reconnection loop) is only called for the connections that were initially successful.

   * The Problem: If a relay is down or unreachable when the rmb-sdk-go peer application starts, that specific relay will be permanently excluded from the peer's active connection pool for  the entire lifetime of that peer instance. Its dedicated reconnection goroutine will never be launched.

   * When it would be retried (currently): It would not be automatically retried. The only way for the peer to attempt connecting to that relay again would be for the entire Go application to be restarted.

  Impact:

   * Reduced Redundancy: If a configured relay is temporarily down or unreachable when the Go peer starts, that relay will never be retried or connected to for the entire lifetime of the peer instance. This significantly reduces the peer's intended redundancy and resilience.
   * Potential Message Loss: While the peer updates its on-chain twin information with the actual connected relays, there's a window where other peers might attempt to route messages via a relay that this peer intended to use but failed to connect to at startup.

  Suggested Fix:

  The Peer should manage all configured relay URLs from the outset. For every configured relay, a dedicated InnerConnection instance should be created, and its Start() method (containing the persistent reconnection logic) should be launched. This ensures that all intended relays are continuously attempted until a connection is established, making the peer truly resilient.

PR #1392, opened 2025-07-23
Ensure relays that fail at startup are retried

sameh-farouk:
### Description
**Problem Context**: During peer startup, only relays that connect successfully have their reconnection goroutine (InnerConnection.Start()) started. Relays that fail the initial connection are ignored forever.
**Impact**: If a relay is down at startup, it is never retried unless the app is restarted. This reduces redundancy and resilience.

This pull request (PR) introduces a change in behavior for the SDK. By default, the SDK will now automatically attempt to self-heal by retrying all configured relays in the background, even if any are down at startup.

An option `WithRequireFunctionalRelayOnStartup` was added to allow applications to choose stricter startup checks if needed. If it is enabled, the SDK will require at least one relay to be functional at startup. If no relays are functional, the SDK will return an error and will not start, which reflects the old behavior. This is not enabled by default. 

### Changes

- Refactor `getRelayConnections`:
  - Creates `InnerConnection` instances for valid relay URLs (those that can be parsed).
  - Invalid URLs are logged and skipped (no retry, no goroutine).
  - The function returns all valid connections, regardless of initial connection success.

- Refactor `NewPeer`:
  - Starts the reconnection goroutine `Start` for all valid InnerConnection instances, ensuring that even relays that are temporarily down at startup will be retried.
  - There is no longer any logic that filters out relays based on initial connection success.

### Related Issues

https://github.com/threefoldtech/tfgrid-sdk-go/issues/1389

### Checklist

- [ ] Tests included
- [X] Build pass
- [ ] Documentation
- [X] Code format and docstring

### DIFF
--- a/rmb-sdk-go/peer/peer.go
+++ b/rmb-sdk-go/peer/peer.go
@@ -37,6 +37,13 @@ type Handler func(ctx context.Context, peer *Peer, env *types.Envelope, err erro
 
 type cacheFactory = func(inner TwinDB, chainURL string) (TwinDB, error)
 
+var (
+	// ErrNoValidRelayURLs is returned when no valid relay URLs are provided.
+	ErrNoValidRelayURLs = errors.New("no valid relay URLs provided")
+	// ErrNoFunctionalRelayAtStartup is returned when no relay connections are functional at startup.
+	ErrNoFunctionalRelayAtStartup = errors.New("no relay connections are functional at startup")
+)
+
 type peerCfg struct {
 	// Require at least one working relay at startup (default: false, for backward compatibility)
 	RequireFunctionalRelayOnStartup bool
@@ -157,7 +164,7 @@ func getRelayConnections(relayURLs []string, identity substrate.Identity, sessio
 	}
 
 	if len(connections) == 0 {
-		return nil, nil, errors.New("no valid relay URLs provided")
+		return nil, nil, ErrNoValidRelayURLs
 	}
 
 	sort.Slice(validRelayURLs, func(i, j int) bool {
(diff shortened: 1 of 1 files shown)


rmb-sdk-go/peer/peer.go
```diff
@@ -130,38 +140,32 @@ func generateSecureKey(identity substrate.Identity) (*secp256k1.PrivateKey, erro
...
-		}
-
-		log.Warn().Str("url", relayURL).Msg("failed to connect")
+		connections = append(connections, conn)
 	}
 
-	if len(successfulRelayURLs) == 0 {
-		return nil, nil, errors.New("failed to connect to any relay")
+	if len(connections) == 0 {
+		return nil, nil, errors.New("no valid relay URLs provided")
```
xmonader: extract static errors into variables

rmb-sdk-go/peer/peer.go
```diff
@@ -260,6 +264,20 @@ func NewPeer(
 		return nil, err
 	}
 
+	// Hybrid approach: if RequireFunctionalRelayOnStartup is true, require at least one working relay at startup
+	if cfg.RequireFunctionalRelayOnStartup {
+		hasWorking := false
```
xmonader: you can replace the logic from 269 to 279 with IndexFun

```
	if cfg.RequireFunctionalRelayOnStartup {
		firstWorking := slices.IndexFunc(conns, func(c InnerConnection) bool {
			return c.TryConnect()
		})
		if firstWorking == -1 {

			....
        }
```

rmb-sdk-go/peer/peer.go
sameh-farouk: I have extracted the static error strings touched in this PR into package-level variables

rmb-sdk-go/peer/peer.go
sameh-farouk: This is a good use of `slices.IndexFunc`
Done!

xmonader (approved):

rmb-sdk-go/peer/peer.go
```diff
@@ -63,6 +72,14 @@ func WithEncryption(enable bool) PeerOpt {
 }
 
 // WithRelay set up the relay url, default is mainnet relay
```
rawdaGastan: I think this comment should be moved to withRelay function

rmb-sdk-go/peer/peer.go
```diff
@@ -260,6 +271,16 @@ func NewPeer(
 		return nil, err
 	}
 
+	// Hybrid approach: if RequireFunctionalRelayOnStartup is true, require at least one working relay at startup
+	if cfg.RequireFunctionalRelayOnStartup {
```
rawdaGastan, line 275: I think old approach requires a valid and working connection to start?

sameh-farouk: We are still parsing URLs as before to ensure their validity. However, the strict requirement of having a working relay at startup is now configurable with that falg and disabled by default. If all relays are down, the system will start and continue making connection attempts.

rawdaGastan (approved):

Issue #1393, opened 2025-07-27
depend on external config source (zos-config)


Omarabdul3ziz:
depending on external config source like zos depending on zos-config will make it easier to track dependencies, mabye we can use the same repo renamed for clearer setup

SalmaElsoly:
### work completed:
- fetch remote configs

### wip:
- still in testing

xmonader:
Pause working on this please @SalmaElsoly

PR #1423, opened 2025-09-01
fix: update network URLs

Omarabdul3ziz:
until we do https://github.com/threefoldtech/tfgrid-sdk-go/issues/1393


sameh-farouk (changes requested):
Please deduplicate URLs; let's stick to `/ws` variant

xmonader (approved):

Issue #571, opened 2023-12-26
ratelimit unprotected calls on the farmerbot


xmonader:
check https://github.com/threefoldtech/tfgrid-sdk-go/pull/509#discussion_r1436152246

PR #1447, opened 2025-10-20
rate limit farmerbot find node calls

rawdaGastan:
### Related Issues

- https://github.com/threefoldtech/tfgrid-sdk-go/issues/571

### DIFF
--- a/farmerbot/internal/farmerbot.go
+++ b/farmerbot/internal/farmerbot.go
@@ -17,6 +17,8 @@ import (
 	"github.com/threefoldtech/tfgrid-sdk-go/rmb-sdk-go/peer"
 )
 
+const defaultMaxGPUItemsToFind = uint32(10)
+
 // FarmerBot for managing farms
 type FarmerBot struct {
 	*state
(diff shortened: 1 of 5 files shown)


farmerbot/internal/farmerbot.go
```diff
@@ -153,12 +157,29 @@ func (f *FarmerBot) serve(ctx context.Context) error {
...
 		if err := json.Unmarshal(payload, &options); err != nil {
 			return nil, fmt.Errorf("failed to load request payload: %w", err)
 		}
 
+		maxGPUItemsToFind := defaultMaxGPUItemsToFind
+		if f.config.MaxGPUsToFind > 0 {
+			maxGPUItemsToFind = f.config.MaxGPUsToFind
+		}
+
+		if len(options.GPUVendors) > int(maxGPUItemsToFind) {
```
Omarabdul3ziz, line 176: could you pls clarify why we need to limit this?

farmerbot/internal/farmerbot.go
```diff
@@ -153,12 +157,29 @@ func (f *FarmerBot) serve(ctx context.Context) error {
 	})
 
 	nodeRouter.WithHandler("findnode", func(ctx context.Context, payload []byte) (interface{}, error) {
+		allowed, waitTime := f.findNodeLimiter.Allowed()
```
Omarabdul3ziz, line 160: should we add rate-limiting to other calls as well?

farmerbot/internal/farmerbot.go
rawdaGastan: because we need to control the items we should found instead of letting user provide a very large list of vendors

farmerbot/internal/farmerbot.go
rawdaGastan: this is the only code can be used from users other calls are used from farmer only

Omarabdul3ziz (approved):

PR #1459, opened 2025-11-03
implement idempotent deployment creation by reusing existing contracts

Omarabdul3ziz:
- https://github.com/codescalers/kubecloud/issues/739
- https://github.com/threefoldtech/tfgrid-sdk-go/issues/1462


AbdelrahmanElawady:
can we rely on errors returned from both the chain and the zos node to determine if it's already deployed or not instead of sending extra requests to both tfchain and zos?
zos reports if the deployment already exists with this error: https://github.com/threefoldtech/zosbase/blob/main/pkg/provision/interface.go#L58
maybe tfchain also has a similar error?
also the case if a contract is created but the client terminated before node deployment is not handled here I think

Omarabdul3ziz:
found this for chain https://github.com/threefoldtech/tfchain/blob/b950278ead307174e560471a8903cb7cce743e71/substrate-node/pallets/pallet-smart-contract/src/grid_contract.rs#L70

for the second part. checking on the node will make some orphan contracts which will be out of deployer hand to free later also node takes sometime to start the workload.

but definitely we need to make sure at least node received request of deployment

AbdelrahmanElawady:
the time for the workload to start doesn't affect the response that the deployment is already sent, right?

grid-client/deployer/deployer.go
```diff
@@ -156,27 +156,48 @@ func (d *Deployer) deploy(
...
 
 			dl.ContractID = contractID
+			if err != nil && strings.Contains(err.Error(), "ContractIsNotUnique") {
+				// Contract not unique: find existing contract using hash
+				existingContractID, found := d.findExistingOnChainContract(node, hash)
+				if !found {
+					return currentDeployments, errors.Wrapf(err, "failed to create contract on node %d and could not find existing contract", node)
+				}
+				dl.ContractID = existingContractID
+			}
```
AbdelrahmanElawady:
```suggestion
			if strings.Contains(err.Error(), "ContractIsNotUnique") {
				contractID, found = d.findExistingOnChainContract(node, hash)
				if !found {
					return currentDeployments, errors.Wrapf(err, "failed to create contract on node %d and could not find existing contract", node)
				}
			}  else if err != nil {
				return currentDeployments, errors.Wrapf(err, "failed to create contract on node %d", node)
			}
			dl.ContractID = contractID
```

grid-client/subi/substrate_manager.go
```diff
@@ -153,6 +154,15 @@ func (s *SubstrateImpl) CreateNodeContract(identity substrate.Identity, node uin
 	return res, normalizeNotFoundErrors(err)
 }
 
+// GetContractWithHash gets a contract by hash
+func (s *SubstrateImpl) GetContractWithHash(identity substrate.Identity, node uint32, hash []byte) (uint64, error) {
+	s.m.Lock()
+	defer s.m.Unlock()
```
AbdelrahmanElawady: no need to lock here it's a query for an existing contract not an extrinsic

grid-client/deployer/deployer.go
```diff
@@ -156,27 +156,48 @@ func (d *Deployer) deploy(
...
+
+			// Update deployment with contract ID and send to node
 			err = client.DeploymentDeploy(ctx, dl)
 			if err != nil {
-				rerr := d.substrateConn.EnsureContractCanceled(d.identity, contractID)
+				// If deployment exists, continue as already deployed
+				if strings.Contains(err.Error(), "ErrDeploymentExists") || strings.Contains(err.Error(), "deployment exists") {
```
AbdelrahmanElawady: I don't think zos returns `ErrDeploymentExists`

AbdelrahmanElawady: the error is just `exists`, right?
https://github.com/threefoldtech/zosbase/blob/main/pkg/provision/interface.go#L58

grid-client/deployer/deployer.go
```diff
@@ -156,27 +156,48 @@ func (d *Deployer) deploy(
...
+					currentDeployments[node] = dl.ContractID
+					log.Debug().
+						Uint32("node", node).
+						Uint64("contractID", dl.ContractID).
+						Msg("deployment exists, continuing as already deployed")
+					continue
```
AbdelrahmanElawady: this continue without checking that the workload state is Ok

grid-client/deployer/deployer.go
```diff
@@ -258,6 +279,30 @@ func (d *Deployer) deploy(
 	return currentDeployments, nil
 }
 
+// findExistingOnChainContract tries to find an existing active contract on-chain for a given node
+// that matches the provided deployment hash and is owned by the current twin.
+func (d *Deployer) findExistingOnChainContract(node uint32, hash []byte) (uint64, bool) {
+	if len(hash) == 0 {
+		return 0, false
+	}
```
AbdelrahmanElawady: it can never be 0, right?

grid-client/deployer/deployer.go
```diff
@@ -258,6 +279,30 @@ func (d *Deployer) deploy(
...
+	if err != nil {
+		return 0, false
+	}
+
+	// Verify the contract is valid and owned by current twin
+	contract, err := d.substrateConn.GetContract(contractID)
+	if err != nil {
+		return 0, false
+	}
+	if !contract.IsCreated() || contract.TwinID() != d.twinID {
```
AbdelrahmanElawady: if contracts match hash then twins are identical, right?

grid-client/deployer/deployer.go
```diff
@@ -258,6 +279,30 @@ func (d *Deployer) deploy(
 	return currentDeployments, nil
 }
 
+// findExistingOnChainContract tries to find an existing active contract on-chain for a given node
+// that matches the provided deployment hash and is owned by the current twin.
+func (d *Deployer) findExistingOnChainContract(node uint32, hash []byte) (uint64, bool) {
```
AbdelrahmanElawady: prefer if it returns errors for more visibility

AbdelrahmanElawady (changes requested):

AbdelrahmanElawady:
also I think if we have errors defined in node/chain clients would be better instead of matching strings, if it's possible

grid-client/deployer/deployer.go
```diff
@@ -155,28 +156,49 @@ func (d *Deployer) deploy(
...
 			dl.ContractID = contractID
+
+			// Update deployment with contract ID and send to node
 			err = client.DeploymentDeploy(ctx, dl)
 			if err != nil {
-				rerr := d.substrateConn.EnsureContractCanceled(d.identity, contractID)
-				if rerr != nil {
-					return currentDeployments, errors.Wrapf(err, "error cancelling contract: %s; you must cancel it manually (id: %d)", rerr, contractID)
+				// If deployment exists, continue as already deployed
+				if !errors.Is(err, provision.ErrDeploymentExists) {
```
AbdelrahmanElawady: I don't think the errors will match since the first error is received from node client as rmb message not the same reference as provision.ErrDeploymentExists

grid-client/deployer/deployer.go
```diff
@@ -155,28 +156,49 @@ func (d *Deployer) deploy(
...
 
 			contractID, err := d.substrateConn.CreateNodeContract(d.identity, node, dl.Metadata, hashHex, publicIPCount, newDeploymentSolutionProvider[node])
-			log.Debug().Uint64("CreateNodeContract returned id", contractID)
 			if err != nil {
-				return currentDeployments, errors.Wrapf(err, "failed to create contract on node %d", node)
+				if strings.Contains(err.Error(), "ContractIsNotUnique") {
+					contractID, err = d.substrateConn.GetContractWithHash(d.identity, node, hash)
+					if err != nil {
+						return currentDeployments, errors.Wrapf(err, "failed to find existing contract on node %d", node)
+					}
```
AbdelrahmanElawady: I think a check that the contract is Ok should be done in case it matched with a cancelled contract

grid-client/deployer/deployer.go
```diff
@@ -154,29 +154,58 @@ func (d *Deployer) deploy(
...
+			// Update deployment with contract ID and send to node
+			err = nodeClient.DeploymentDeploy(ctx, dl)
 			if err != nil {
-				rerr := d.substrateConn.EnsureContractCanceled(d.identity, contractID)
-				if rerr != nil {
-					return currentDeployments, errors.Wrapf(err, "error cancelling contract: %s; you must cancel it manually (id: %d)", rerr, contractID)
+				// If deployment exists, continue as already deployed
+				if !contractRused || !strings.Contains(err.Error(), "exists") {
+					// Other deployment error: cancel contract
+					rerr := d.substrateConn.EnsureContractCanceled(d.identity, dl.ContractID)
```
rawdaGastan, line 189: isn't it dangerous here? we can cancel existing ones (will the error check be enough?)
should we check ownership first?

grid-client/deployer/deployer.go
```diff
@@ -154,29 +154,58 @@ func (d *Deployer) deploy(
 			}
 			log.Debug().Uint32("Number of public ips", publicIPCount)
 
+			contractRused := false
 			contractID, err := d.substrateConn.CreateNodeContract(d.identity, node, dl.Metadata, hashHex, publicIPCount, newDeploymentSolutionProvider[node])
-			log.Debug().Uint64("CreateNodeContract returned id", contractID)
 			if err != nil {
-				return currentDeployments, errors.Wrapf(err, "failed to create contract on node %d", node)
+				if !strings.Contains(err.Error(), "ContractIsNotUnique") {
```
rawdaGastan, line 160: can't we use errors not string?

rawdaGastan (commented):
please add an issue in the repo for the pr (we can mention the kubecloud one)

grid-client/deployer/deployer.go
AbdelrahmanElawady: chain client doesn't have this error defined. this is coming directly from the chain

grid-client/deployer/deployer.go
AbdelrahmanElawady: we can't cancel a contract we don't own

rawdaGastan (approved):

AbdelrahmanElawady (approved):

PR #1475, opened 2025-12-02
Fix rmb client peer process panic

sameh-farouk:
### Description

This PR fix a bug where two goroutines shared responsibility for the same channel:
`Call`: closed it,
`router`: sent to it,
without synchronization around close.

### Changes

- only delete the map entry; never close the per‑call channel.
That makes it impossible for `router` to hit a closed channel, while still allowing late responses to be ignored safely.

### Related Issues

- #1476

### Checklist

- [ ] Tests included
- [ ] Build pass
- [ ] Documentation

### DIFF
--- a/rmb-sdk-go/peer/rpc.go
+++ b/rmb-sdk-go/peer/rpc.go
@@ -103,8 +103,6 @@ func (d *RpcClient) CallWithSessionWithTags(ctx context.Context, twin uint32, se
 
 	ch := make(chan incomingEnv, 1)
 	defer func() {
-		close(ch)
-
 		d.m.Lock()
 		delete(d.responses, id)
 		d.m.Unlock()


AbdelrahmanElawady:
I see the channel referenced in two places [here](https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/rmb-sdk-go/peer/rpc.go#L121-L126) and [here](https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/rmb-sdk-go/peer/rpc.go#L80-L83). while currently in both cases freeing the channel won't block, any changes to code to make the channel a sync point will be a leak that will be hard to debug as the receiver will be waiting forever. I think the idiomatic way to do it in Go is to make the writer manage the channel and close it not from the reader

sameh-farouk:
> I see the channel referenced in two places [here](https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/rmb-sdk-go/peer/rpc.go#L121-L126) and [here](https://github.com/threefoldtech/tfgrid-sdk-go/blob/development/rmb-sdk-go/peer/rpc.go#L80-L83). while currently in both cases freeing the channel won't block, any changes to code to make the channel a sync point will be a leak that will be hard to debug as the receiver will be waiting forever.

- the first place you refered to is `CallWithSessionWithTags` which when returns the local variable `ch` goes out of scope, then defer runs and does `delete(d.responses, id)`.

- the second place is `router`, where you see the select takes `case incoming = <-ch:` and receives that value. After the select we left with `incoming` (the struct) holds the response but `ch` is not used anymore in the function.

Result:
- No references to ch remain:
- Not in locals (function has returned),
- Not in d.responses (entry deleted),
- Not in any goroutine (router only ever had a temporary local per call).

>  I think the idiomatic way to do it in Go is to make the writer manage the channel and close it not from the reader

This a good rule when you’re doing ranged reads (keeps receiving until the channel is closed) and using close as a signal.

In our code here there is no long‑running goroutine that doing ranged reads and we don't need close as a signal.
If you tried to follow “writer closes” strictly here, it would hard to synchronize without a lot of extra state. You will see that there is a lot of moving parts just to close a channel that you don't actually need to close and you may end up back in exactly the race we had.

Given that you do not need to close channels just to reclaim memory as GC takes care of that once they’re unreachable (Channels aren't like files; you don't usually need to close them) and with the current code structure, the cleanest and safest approach is exactly what we have here: no one closes this channel, it dies when becomes unreachable.

AbdelrahmanElawady (approved):
Approved as this is causing issues in the current running systems.

PR #1481, opened 2025-12-03
add node provisioning probe

Omarabdul3ziz:
# in progress
to add :
- [ ] handle dedicated/rented nodes
- [ ] deploy/record in batches
- [ ] handle zoslight nodes
- [ ] after each cycle & on startup: delete all contracts - in case something stuck
- [ ] add timescale to the live `processor_db` instead
- [ ] use better queries, `sqlc`?
- [ ] add auto retention based on configured window
- [ ] add more rates to score: healthy? deployment duration?


xmonader:
before proceeding please do the following

1- change the name to nodes-deployment-checker
2- switch to gin
3- use swagger/gin integration
4- also stick to gorm, no need to introduce something new in this project

